Issue #5, opened 2020-07-21
Concept of animating moving


Wikunia:
Besides interpolating between objects as described in #4 we probably want an easy way to move paths from point A to B

Instead of writing something like
```
for i=1:100
  @layer begin
    translate(0, i)
    DRAWING
  end
end
```

we probably want to have a general move function like
```
move(obj, from, to, frames, easingfunction)
```

Wikunia:
Needs something like an easingfunction but in general this works now.

PR #6, opened 2020-07-21
example version of a simple animating move

Wikunia:
Some ideas for #5 

```
ground(scene, framenumber) = background("white")

function latex_text()
    sethue("black")
    latex(L"\begin{equation}\left[\begin{array}{cc}  2 & 3 \\  4 & \sqrt{5} \\  \end{array} \right] \end{equation}", 50)
end

demo = Movie(500, 500, "test", 1:100)
from = Point(-200, -200)
to = Point(-20, -130)
from_rot = 0.0
to_rot = π/8
animate(demo, [
    Scene(demo, ground, 1:100)
    Scene(demo, transition(latex_text, Rotation(from_rot, to_rot), Translation(from, to)), 1:50)
    Scene(demo, transition(latex_text, Rotation(to_rot, from_rot), Translation(to, from)), 51:100)
], creategif=true, pathname="latex.gif")
```

creates:
latex

It uses the standard Luxor syntax for animation. New is the `transition` function which takes a function that gets called as well as a list of transitions. 
Based on the frame number the transition is called. This works nicely (for this toy example at least) as an abstraction of the luxor way or "hardcoding" it. 

**Next:**
- [ ] Test cases
- [ ] Easing functions

You might wanna have a look at it @TheCedarPrince


TheCedarPrince:
This is great -- I will see if I can mess around with this to produce line objects.

Wikunia:
I thought about it again @TheCedarPrince and I think we need some more expressiveness :smile: 
For example a `Scene` wasn't able to return something. 
A new structure can be found in the latest commit:

## Action
Similar to the Luxor Scene with a bit more freedom. 
It takes:
- frame range
- name which can be used in other actions to refer to it
- drawing function
- Transitions

## Transition
At the moment `Translation` and `Rotation` needs `Scale` and probably some form of a color transition.
Each of these objects can hold the information directly like move from the origin to `Point(100, 100)` or it uses a symbol 
from a previous `action` which might be more clear in the example.

```
function ground(args...) 
    background("white")
    sethue("black")
end

function circ(p=O, color="black")
    sethue(color)
    circle(p, 25, :fill)
    return Transformation(p, 0.0)
end

function anim()
    from = Point(-200, -200)
    to = Point(-20, -130)
    p1 = Point(0,-100)
    p2 = Point(0,-50)
    from_rot = 0.0
    to_rot = 2π

    demo = Video(500, 500)
    javis(demo, [
        Action(1:100, ground),
        Action(1:100, :red_ball, (args...)->circ(p1, "red"), Rotation(from_rot, to_rot)),
        Action(1:100, (args...)->circ(p2, "blue"), Rotation(to_rot, from_rot, :red_ball))
    ], tempdirectory="test/current/images", creategif=true, pathname="test/current/dancing_circles.gif")
end
```

creates:
dancing_circles

here the blue ball rotates around the red ball by using`:red_ball` as the center of the rotation.

Wikunia:
I think I'll add easing functions in a different commit so ready to review @TheCedarPrince whenever you like :smile:

Wikunia:
I'll merge this once the tests pass. Then we might need to check on other PRs if we need to rebase or whether git handles everything. The PRs should be either to review then.

Issue #4, opened 2020-07-21
Interpolate between two paths


Wikunia:
Would be nice to have a way to convert from a circle to a square like described here:
https://www.youtube.com/watch?v=ABB1qqbesPM
and https://github.com/JuliaGraphics/Luxor.jl/blob/1cdcb74fdfffde264c3f8dfacb84a84e0085f154/docs/src/simplegraphics.md

but for every path to every other path such that 
`x^2 + y^2 = r^2`

can be transformed into a circle somehow :joy: 
That might look ridiculous but there is only one way to figure it out.

**Just some links:**
How this can be done in d3.js:
https://github.com/veltman/flubber
https://www.youtube.com/watch?v=PLc1y-gim_0

Wikunia:
This is probably out of scope in the beginning but maybe we can have a simple version which can transform one polygon to another. Thoughts @TheCedarPrince ?

Wikunia:
Next up is: #80

Wikunia:
We'll create smaller subissues like #80 from now on

PR #15, opened 2020-07-24
Simple morph between paths. SORRY FOR THE WEIRD COLORS

Wikunia:
First implementation of #4 
This currently works only for objects with the same number of vertices or a multiples. :smile: 
star2rectangle

**Todo:** will be done in subsequent PRs :smile: 
- [ ] Ability to choose `:fill` or `:stroke`
- [ ] How to handle bigger lines? Do we have access to it?


Wikunia:
Some more examples which are not that colorful :smile: 

star2circle
star2rectangle

Wikunia:
Oh maybe there was a misunderstanding. I thought we have full comments for exported functions and maybe a bit less information for private functions. Additionally explanation of `Video` and `Action` once for this main function `javis` and referring to it from other functions to not explain `video, action, frame` over and over again. Does that make sense?

TheCedarPrince:
Ah that makes a lot of sense. I think there was a slight misunderstanding that I took to mean we only document kwargs for every function. I like this approach better. I am going to make a note of that in `contributing.md`.

PR #18, opened 2020-07-26
[WIP] Grid Lines Animation

TheCedarPrince:
Hey @Wikunia ,

Working on creating the gridlines animation. So far, I have the idea of what I want in my head being:

1. A function that calculates how many horizontal and vertical grid lines to make based on canvas size
2. Maybe making a mutable line object as opposed to the `line` object provided by `Luxor.jl`
3. Creation of zero-lines function dependent on canvas size
4. Creating a function that can "grow" the lines for the grid
5. Hooking into `FFMPEG`

Right now, I generated a gif with at least the basic idea of drawing a line:

grid_line

Not very thrilling yet and it is not using `javis` just yet as I am unsure of how to intelligently use it. More to come.


Wikunia:
Can you clarify what you mean by your second point of  creating a mutable Line object? We should always make sure that we have different names then Luxor to not have ambiguity issues.
You probably want to have a look at the javis or the animate function in Luxor for ffmpeg

Wikunia:
BTW your javis try failed because you tried to define two functions. The javis function only accepts a single function and then a couple of transformation structs if needed. The single function can take in the arguments movie, action and frame such that you in general have all the information needed to do whatever you like to do in this animation action if that makes sense.

Out of the top of my head without testing I would say that this will work:
```
javis(video,
Action(1:100, :line, line_move)
)

function line_move(video, action, frame)
    t = (frame-first(action.frames))/(length(action.frames)-1)
    line(Point(-200,0), Point(-200,0)+t*(Point(200,0)-Point(-200,0)), :stroke)
end
```

or something like that. This is of course the user unfriendly version which needs to be abstracted into an easy to use version for the grid animation.

TheCedarPrince:
I am going to continue hacking on this but we are making progress! We have a proof of concept gif!

grid_test

@Wikunia , I tried your proposed fix about attempting to change this code snippet in `Javis.jl`:

```julia
    filecounter = 1
    for frame in frames
        Drawing(video.width, video.height, "$(tempdirectory)/$(lpad(filecounter, 10, "0")).png")
        origin()
        start_translation = Point(gettranslation()...)
        # this frame needs doing, see if each of the scenes defines it
        for action in actions
            if frame in action.frames
                @layer begin
                    compute_transformation!(action, video, frame)
                    perform_transformation(action)
                    res = action.func(video, action, frame)
                    if res isa Point
                        trans = cairotojuliamatrix(getmatrix())*res
			trans.p -= start_translation
                        video.defs[action.id] = trans
                    end
                end
            end
        end
        finish()
        filecounter += 1
    end
```

However, though I was able to create the line gif above, I could not get all the tests to pass. Here is the error message from the testing:

```
Test Summary: | Pass  Total
O(logn)       |    1      1
┌ Warning: test fails because PSNR -19.72929573059082 < 25
└ @ ReferenceTests ~/.julia/packages/ReferenceTests/JDrHL/src/equality_metrics.jl:45
Test for "dancing_circles_16.png" failed.
Dancing circles: Error During Test at /home/src/Projects/javis/test/animations.jl:23
  Got exception outside of a @test
  You need to run the tests interactively with 'include("test/runtests.jl")' to update reference images
  Stacktrace:
   [1] error(::String) at ./error.jl:33
   [2] test_reference(::FileIO.File{FileIO.DataFormat{:PNG}}, ::Array{RGB{Normed{UInt8,8}},2}, ::Nothing, ::Nothing; kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/cedarprince/.julia/packages/ReferenceTests/JDrHL/src/test_reference.jl:140
   [3] test_reference(::FileIO.File{FileIO.DataFormat{:PNG}}, ::Array{RGB{Normed{UInt8,8}},2}, ::Nothing, ::Nothing) at /home/cedarprince/.julia/packages/ReferenceTests/JDrHL/src/test_reference.jl:100
   [4] test_reference(::String, ::Array{RGB{Normed{UInt8,8}},2}; by::Nothing, render::Nothing, kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/cedarprince/.julia/packages/ReferenceTests/JDrHL/src/test_reference.jl:90
   [5] test_reference(::String, ::Array{RGB{Normed{UInt8,8}},2}) at /home/cedarprince/.julia/packages/ReferenceTests/JDrHL/src/test_reference.jl:90
   [6] top-level scope at /home/src/Projects/javis/test/animations.jl:41
   [7] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
   [8] top-level scope at /home/src/Projects/javis/test/animations.jl:24
   [9] include(::String) at ./client.jl:439
   [10] top-level scope at /home/src/Projects/javis/test/runtests.jl:8
   [11] include(::String) at ./client.jl:439
   [12] top-level scope at none:6
   [13] eval(::Module, ::Any) at ./boot.jl:331
   [14] exec_options(::Base.JLOptions) at ./client.jl:264
   [15] _start() at ./client.jl:484
  
Test Summary:   | Error  Total
Dancing circles |     1      1
ERROR: LoadError: LoadError: Some tests did not pass: 0 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /home/src/Projects/javis/test/animations.jl:23
in expression starting at /home/src/Projects/javis/test/runtests.jl:8
ERROR: Package Javis errored during testing
```

Although it is obviously not great, I wrapped the line where the error occurs with a try-catch block. Now, all the tests work. I figure while I work on extending the functionality of this animation, this will do while you can see where the error is coming from @Wikunia :


```julia
    filecounter = 1
    for frame in frames
        Drawing(video.width, video.height, "$(tempdirectory)/$(lpad(filecounter, 10, "0")).png")
        origin()
        start_translation = Point(gettranslation()...)
        # this frame needs doing, see if each of the scenes defines it
        for action in actions
            if frame in action.frames
                @layer begin
                    compute_transformation!(action, video, frame)
                    perform_transformation(action)
                    res = action.func(video, action, frame)
                    if action.id !== nothing && res !== nothing
                        trans = cairotojuliamatrix(getmatrix())*res
			try
				trans.p -= start_translation
			catch 
				continue
			end
                        video.defs[action.id] = trans
                    end
                end
            end
        end
        finish()
        filecounter += 1
    end
```

TheCedarPrince:
We got it @Wikunia ! :tada: :tada: :tada:

grid_test

I think this is ready for review now. I'll let you see how to fix the problem in `src/Javis.jl`. Once you make your comments or anything you are thinking about wanting to add, please let me know and then I can start merging it as a core functionality.

Wikunia:
Great! Thanks for pinging me
Some general comments first while I'm on vacation 😊
My suggestion was to use `if action.id !== nothing && res isa Point`
The first part checks whether the action has a name and the second whether the action returned a point if that's the case then save the point to the name.
Additionally currently you haven't integrated it into Javis yet, right? You basically built it using Javis which is an important first step. Next up should be to make it simpler for the user to achieve the same thing and have flexibility as how many lines if not default, speed of animation, maybe direction of animation, etc.

TheCedarPrince:
Hey @Wikunia ,

That fix you proposed worked. All tests are now passing without any hacks going on! :tada:

As for adding additional functionality, I added the ability to draw zero lines. Furthermore, I have now added the ability to choose directions where grid lines and zero lines finish:

### Bottom Right:

grid_test

### Bottom Left:

grid_test

### Top Right:

grid_test

### Top Left:

grid_test

## Current Problem

This is somewhat embarrassing but I cannot seem to add kwargs to the function definition of my `draw_grid` function. 

```julia
using Javis
using Luxor

function grid_animation()

    video = Video(500, 500)
    javis(
        video,
        [
            Action(1:100, ground),
            # Action(1:100, :line, draw_grid(truth_value=true)),
            Action(1:100, :line, draw_grid),
            Action(1:100, :line, zero_lines),
        ],
        tempdirectory = "/home/src/Projects/javis/test/grid_test",
        creategif = true,
        pathname = "/home/src/Projects/javis/test/grid_test.gif",
    )

end

# function draw_grid(video::Video=nothing, action::Action=nothing, frame::Int=1, truth_value::Bool=false)
function draw_grid(video::Video, action::Action, frame::Int)
```

I want to have the two commented lines work by passing in a boolean from the `javis` function, but I cannot get it working. What am I missing? Thanks @Wikunia -- after I get this figured out, I will keep experimenting with this and add it into the `src/Javis.jl` file and will begin developing some tests. :smile:

Wikunia:
Good progress! I know the keyword thing is a bit confusing but it makes sense.
The Action takes a function name but you give it the result of a function as you directly call the function. You can either use `(args...)->draw_line(args..., truth_vale=true)` so define an anonymous function or:
Have a draw line function which takes one argument (the truth value) and returns a function. For this have a look at "closures" in Julia.
Hope this helps.

TheCedarPrince:
Great. That worked @Wikunia ! In the draw line function, why did I have to splat the preceeding args before a kwarg in the definition of an julia anonymous function? I didn't see that in any documentation that I read anywhere. I was super close to having this definition but missed the `(args..., truth_value=true)` portion.

I will go ahead and start adding this core functionality. I was also debating about possibly splitting this into a separate source file that could be imported by `Javis.jl` as I feel like the main source file is getting a bit too big. What do you think? Should I merge it into the main source file or break it up?

Chat later! Off to sleep :zzz:

Wikunia:
A bit prose about splatting as far as I understand it.
If you have a tuple `t` you can unpack it with `t...` such that `(a,b)` will be `a,b` 
If you have `args...` this means it's the latter but you get the tuple version with `args`. 
This means if you write `args` it's basically one variable and if you use `args...` you have more variables. Hope that makes sense. BTW `args` itself has no meaning other than being a variable. You can also call it `jacob`. It's just convention. It's similar for keyword arguments but there you have a named tuple and after splatting named variables with something assigned to it.
Hope that makes sense.

Wikunia:
Oh and yes please make a new file whenever you think it's helpful. Maybe one can even make one for `latex` which is currently in Javis.jl. I like to keep that one maybe for types and the `javis` method or something like that.

test/grid_animation.jl
```diff
@@ -0,0 +1,158 @@
...
+        tempdirectory = "/home/src/Projects/javis/test/grid_test",
+        creategif = true,
+        pathname = "/home/src/Projects/javis/test/grid_test.gif",
+    )
+
+end
+
+function draw_grid(video::Video, action::Action, frame::Int; grid_direction::Symbol = :TR)
+
+    min_width = video.width / -2
```
Wikunia, line 23: You might want to use `div` instead of `/`

test/grid_animation.jl
```diff
@@ -0,0 +1,158 @@
...
+    max_width = video.width / 2
+
+    min_height = video.height / -2
+    max_height = video.height / 2
+
+    if grid_direction == :BR
+
+        for y_point = min_width:25:max_width
```
Wikunia, line 31: The 25 should be a variable I guess

test/grid_animation.jl
```diff
@@ -0,0 +1,158 @@
...
+            Action(1:100, :line, (args...) -> zero_lines(args..., line_direction = :TL)),
+        ],
+        tempdirectory = "/home/src/Projects/javis/test/grid_test",
+        creategif = true,
+        pathname = "/home/src/Projects/javis/test/grid_test.gif",
+    )
+
+end
+
+function draw_grid(video::Video, action::Action, frame::Int; grid_direction::Symbol = :TR)
```
Wikunia, line 21: For porting it into Javis itself you might want to use `Val` for dispatch here instead of having an `if` `else` for grid_direction. Just for an easier read (at least for me)

test/grid_animation.jl
```diff
@@ -0,0 +1,158 @@
...
+
+function zero_lines(video::Video, action::Action, frame::Int; line_direction::Symbol = :TR)
+
+    min_width = video.width / -2
+    max_width = video.width / 2
+
+    min_height = video.height / -2
+    max_height = video.height / 2
+
+    setline(10)
```
Wikunia, line 106: Should be variable for the user later but I guess you know that.

TheCedarPrince:
Hey @Wikunia opening this for review right now as I think it is ready for an initial merge to give basic functionality. Do you want to go ahead and merge and have me extend its features or would you prefer it to have my features as of now?

test/grid_animation.jl
TheCedarPrince: Definitely agree. Will work on this either now or later in another PR.

test/grid_animation.jl
TheCedarPrince: Yup!

src/Javis.jl
```diff
@@ -2,6 +2,8 @@ module Javis
 
 using Luxor, LaTeXStrings
 
+include("backgrounds.jl")
```
Wikunia, line 5: this should probably be done after the type definitions as you have `::Video` in there for example which is unknown at that point. Probably want to move t to the point where I have the other `include`

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+    background("white")
+    sethue("black")
+end
+
+video = Video(500, 500)
+javis(
+      video,
+      [
+          Action(1:100, ground),
+          Action(1:100, (args...) -> draw_grid(args..., grid_direction = :TL)),
```
Wikunia, line 31: The user would prefer to write `draw_grid(direction=:TL)` I assume. So using a `close` is probably the way to go.
Maybe have a look at #15 
(I would use `direction` as it's inside the `grid` function already.)

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+
+    min_width = div(video.width, -2, RoundDown)
+    max_width = div(video.width, 2, RoundUp)
+
+    min_height = div(video.height, -2, RoundDown)
+    max_height = div(video.height, 2, RoundUp)
+
+    if grid_direction == :BR
+
+        for y_point = min_width:25:max_width
```
Wikunia, line 51: Before merge I think we want the `25` to be an argument.

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+            step = (frame - first(action.frames)) / (length(action.frames) - 1)
```
Wikunia, line 52: You have this line at several places. I think it can be done before you have `if grid_direction`.

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+            start_point = Point(min_width, y_point)
+            end_point = start_point + step * (Point(max_width, y_point) - start_point)
+            line(start_point, end_point, :stroke)
```
Wikunia, line 55: Same as with `step` but you want to have it at the very end.

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+"""
+function zero_lines(video::Video, action::Action, frame::Int; line_direction::Symbol = :TR)
+
+    min_width = div(video.width, -2, RoundDown)
+    max_width = div(video.width, 2, RoundUp)
+
+    min_height = div(video.height, -2, RoundDown)
+    max_height = div(video.height, 2, RoundUp)
+
+    setline(10)
```
Wikunia, line 166: Easy to make an argument for this. I actually would like to have a different color for this maybe but this can be done in a later PR. The `10` should be an argument in this PR I think.

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+
+    step = (frame - first(action.frames)) / (length(action.frames) - 1)
```
Wikunia, line 168: hah here you did what I mentioned above :smile:

src/backgrounds.jl
```diff
@@ -0,0 +1,212 @@
...
+
+    if line_direction == :BR
+
+        start_point = Point(min_width, 0)
+        end_point = start_point + step * (Point(max_width, 0) - start_point)
+        line(start_point, end_point, :stroke)
```
Wikunia, line 174: this again after the `if`s

src/Javis.jl
```diff
@@ -392,6 +394,7 @@ function javis(
     nothing
 end
 
+export draw_grid, zero_lines
```
Wikunia, line 397: Can/Should be done in `backgrounds.jl` I think.

test/grid_animation.jl
Wikunia: Now is preferred as also mentioned in my next review comments.

src/backgrounds.jl
TheCedarPrince: Fixed

src/Javis.jl
TheCedarPrince: Ah. That fixed an error I was encountering. Everything is good now.

src/backgrounds.jl
TheCedarPrince: Yea, I realized the improvement half-way on but forgot to change it in the other scenario. Fixed.

src/Javis.jl
TheCedarPrince: Oh! Didn't know that by including it in the main package, the export would still be executed. Done!

src/backgrounds.jl
TheCedarPrince: Fixed.

src/backgrounds.jl
TheCedarPrince: Yup! I caught this in another part of the code but sadly did not apply it to all parts. :P

src/backgrounds.jl
TheCedarPrince: Hey little confused by the `this` you are referring to. What do you want me to fix here? @Wikunia

Wikunia: The `line` call is the same in every if so move it outside 😊

src/backgrounds.jl
TheCedarPrince: I personally would like our own function to define a line's characteristics. I saw that you are developing a `Line` struct so maybe its on its way? I also saw this conversation so I am excited to see what comes next #23 

Either way, fixed here.

Wikunia: Why world you like to have a different function for this? To be able to get the current width?
The `Line` struct is basically the definition of a geometric line by two points which is used to calculate the projection of a point onto a line for example.

src/backgrounds.jl
Wikunia: More concretely: https://github.com/Wikunia/Javis.jl/blob/7b63527d9b56c463bc74c8d0636d6cbaeb5a754f/src/Javis.jl#L323

src/backgrounds.jl
Wikunia: Maybe mark this as resolved once you pushed your changes. :smile:

src/backgrounds.jl
TheCedarPrince: Ah true. Implemented!

src/backgrounds.jl
TheCedarPrince: My first closure! Always wanted to use one of these things. :smile: Done!

TheCedarPrince:
Hey @Wikunia , I believe this is fully ready to merge. I addressed all your prior points, added in some tests and reference images, made sure all the checks passed, improved the documentation and syntax, and applied the optimizations you suggested.

Think it is good to go? When you give me the OK, I'll merge it! :smile:

src/backgrounds.jl
```diff
@@ -0,0 +1,216 @@
...
+    background("white")
+    sethue("black")
+end
+
+video = Video(500, 500)
+javis(
+      video,
+      [
+          Action(1:100, ground),
+    	  Action(1:100, :line, (video, action, frame)-> draw_grid(video, action, frame, direction = :TL, line_gap = 25)),
```
Wikunia, line 32: I thought you're using a closure now. :stuck_out_tongue:

src/backgrounds.jl
Wikunia: `line` should be after the `if` as you repeat it 8 times.

src/backgrounds.jl
```diff
@@ -0,0 +1,216 @@
...
+    background("white")
+    sethue("black")
+end
+
+video = Video(500, 500)
+javis(
+      video,
+      [
+          Action(1:100, ground),
+	  Action(1:100, :line, (video, action, frame)-> zero_lines(video, action, frame, direction = :TL, line_thickness = 10)),
```
Wikunia, line 150: also a closure as this is very hard to write for the user.

src/backgrounds.jl
```diff
@@ -0,0 +1,216 @@
...
+    min_height = div(video.height, -2, RoundDown)
+    max_height = div(video.height, 2, RoundUp)
+
+    setline(line_thickness)
+
+    step = (frame - first(action.frames)) / (length(action.frames) - 1)
+
+    if direction == :BR
+
+        start_point_x_point = Point(min_width, 0)
```
Wikunia, line 180: This should be `start_point_x` I believe

src/backgrounds.jl
```diff
@@ -0,0 +1,216 @@
...
+        start_point_x = Point(max_width, 0)
+        end_point_x = start_point_x + step * (Point(min_width, 0) - start_point_x)
+
+        start_point_y = Point(0, max_height)
+        end_point_y = start_point_y + step * (Point(0, min_height) - start_point_y)
+
+    end
+
+        line(start_point_x, end_point_x, :stroke)
+        line(start_point_y, end_point_y, :stroke)
```
Wikunia, line 213: Probably want to fix indentation.

src/backgrounds.jl
```diff
@@ -0,0 +1,216 @@
...
+    max_width = div(video.width, 2, RoundUp)
+
+    min_height = div(video.height, -2, RoundDown)
+    max_height = div(video.height, 2, RoundUp)
+
+    step = (frame - first(action.frames)) / (length(action.frames) - 1)
+
+    if direction == :BR
+
+        for y_point = min_width:line_gap:max_width
```
Wikunia, line 60: In general I would prefer using `Val` here as it makes it more extendable.
Maybe someone wants to animate from the origin outwards. At the moment he/she would need to make a PR or use a fork to add another `if` in the future it should be possible to simply create a new function like:
```
function Javis.draw_grid(video, action, frame, ::Val{outwards}; line_gap=25)
```
and call `draw_grid(video, action, frame; direction=:outwards, line_gap=25)`
which dispatches to the new function which might live outside of Javis.
Hope that makes sense. Doesn't need to be done now but is helpful for another PR if you want to work on it.

src/backgrounds.jl
```diff
@@ -0,0 +1,216 @@
...
+      [
+          Action(1:100, ground),
+    	  Action(1:100, :line, (video, action, frame)-> draw_grid(video, action, frame, direction = :TL, line_gap = 25)),
+      ],
+      tempdirectory = "/tmp-directory",
+      creategif = true,
+      pathname = "/tmp-directory/grid_animation.gif"
+     )
+
+```
```
Wikunia, line 39: I think it's quite tedious to write an example like this every time so maybe we want to just give an example:
```
...
 Action(1:100, :line, draw_grid(direction = :TL, line_gap = 25))
...
```
where `...` can be looked up in the `javis` function. (This would be the example after closures are implemented)

Wikunia:
I've added some more comments. For test cases you might want to test both functions in the same movie.
Additionally I aim for good coverage so the other 3 options should be tested as well if possible.

Wikunia:
You have a lot of code reuse such that it might be helpful to think of horizontal lines and of vertical lines separately. 
So instead of 4 "global" if you are better of with 2 ifs for the downward/upward motion and 2 if for left/right motion.

src/backgrounds.jl
TheCedarPrince: Fixed!

src/backgrounds.jl
TheCedarPrince: Done!

src/backgrounds.jl
TheCedarPrince: With the way it is right now, the for loop creates a horizontal line object each iteration in such a way that `start_point` and `end_point` change between each loop. If I were to put the line outside the if, each line would not be made. Am I missing something somewhere?

src/backgrounds.jl
TheCedarPrince: Fair, I agree with that. I liked the idea of having examples in the docstring simply to give a minimal working example of how it is to function. But perhaps this would be better suited to a tutorial (ping #25)

src/backgrounds.jl
TheCedarPrince: That makes sense and was actually something I was thinking about doing, but didn't feel it appropriate for this initial PR. I agree, let's add it in another PR in the future (ping #39).

I am going to resolve this for now.

TheCedarPrince:
Almost done. Just finishing up with the closures!

src/backgrounds.jl
```diff
@@ -0,0 +1,194 @@
...
+
+    else
+
+        for x_point = min_height:line_gap:max_height
+            start_point = Point(x_point, max_height)
+            end_point = start_point + step * (Point(x_point, min_height) - start_point)
+            line(start_point, end_point, :stroke)
+        end
+
+        if occursin("R", direction)
```
Wikunia, line 73: Isn't this whole block the same as in the one under `occursin("B"` ? Additionally `occursin` is probably not the best way to handle this. You might want to ask either `if direction == "TR" || direction == "BR"` which is the most readable or 
`if direction[2] == "R"`

src/backgrounds.jl
```diff
@@ -0,0 +1,194 @@
...
+            end_point_x = start_point_x + step * (Point(min_width, 0) - start_point_x)
+
+        end
+
+    else
+
+        start_point_y = Point(0, max_height)
+        end_point_y = start_point_y + step * (Point(0, min_height) - start_point_y)
+
+        if occursin("R", direction)
```
Wikunia, line 175: same as above :)

src/backgrounds.jl
```diff
@@ -0,0 +1,194 @@
...
+  - `"TL"` - Animation finishes in the **T**op **L**eft corner of the frame.
+  - `"BR"` - Animation finishes in the **B**ottom **R**ight corner of the frame.
+  - `"BL"` - Animation finishes in the **B**ottom **L**eft corner of the frame.
+- `line_gap::Int`: How many pixels between each line. Default: `25`
+
+# Example
+Example call of this function within an `Action`.
+```
+...
+ Action(1:100, :line, draw_grid(direction = "TL", line_gap = 25))
```
Wikunia, line 19: does not work yet as the closure is missing, right?

src/backgrounds.jl
Wikunia: No that's correct. I missed that or rather was wondering to make the loop outside of the if but that is computationally probably not the right thing to do.

Wikunia:
Additionally make sure that you resolve your merge conflicts.

src/backgrounds.jl
TheCedarPrince: We got this sorted out. :smile:

src/backgrounds.jl
```diff
@@ -0,0 +1,164 @@
...
+...
+```
+
+"""
+function draw_grid(; direction::AbstractString = "TR", line_gap = 25)
+
```
Wikunia, line 25: would also remove these white lines

src/backgrounds.jl
```diff
@@ -0,0 +1,164 @@
...
+"""
+function zero_lines(; direction::AbstractString = "TR", line_thickness = 10)
+    return (video, action, frame) -> _zero_lines(
+        video,
+        action,
+        frame;
+        direction = direction,
+        line_thickness = line_thickness,
+    )
```
Wikunia, line 117: this looks totally different than the one in draw_lines. I like this style :wink:

src/backgrounds.jl
```diff
@@ -0,0 +1,164 @@
...
+    min_height = div(video.height, -2, RoundDown)
+    max_height = div(video.height, 2, RoundUp)
+
+    setline(line_thickness)
+
+    # This determines how quickly the animation is drawn
+    step = (frame - first(action.frames)) / (length(action.frames) - 1)
+
+    if direction[1] == 'B'
+	# Top to bottom motion for vertical line
```
Wikunia, line 140: indent the comments as in `draw_lines` or have it consistent at least

Wikunia:
Thanks for the PR. Sorry for the long process but I'm sure it's getting faster the next times. Please feel free to merge this yourself.
I normally like to merge my own PRs when a reviewer accepts them. Same is true when you review my PRs. Sometimes I find myself having a new idea which I want to try before it gets merged :wink: Just some comments from my project management perspective

PR #19, opened 2020-07-28
used Documenter.jl to build documentation page

sudomaze:
Here is my attempt to build the documentation page using Documenter.jl. It seems that @Wikunia did everything required to build the documentation page correctly. I have added README.md (renamed to index.md) and contributing.md into /docs/src. In addition, I have modified make.jl to run my local build without a warning.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -6,17 +6,18 @@ makedocs(;
     authors="Ole Kröger <o.kroeger@opensourc.es> and contributors",
     repo="https://github.com/Wikunia/Javis.jl/blob/{commit}{path}#L{line}",
     sitename="Javis.jl",
-    format=Documenter.HTML(;
-        prettyurls=get(ENV, "CI", "false") == "true",
-        canonical="https://Wikunia.github.io/Javis.jl",
-        assets=String[],
-    ),
+    #format=Documenter.HTML(;
+    #    prettyurls=get(ENV, "CI", "false") == "true",
+    #    canonical="https://Wikunia.github.io/Javis.jl",
+    #    assets=String[],
+    #),
     pages=[
         "Home" => "index.md",
         "References" => "references.md",
+        "Contributing" => "contributing.md",
     ],
 )
 
-deploydocs(;
-    repo="github.com/Wikunia/Javis.jl",
-)
+#deploydocs(;
+#    repo="github.com/Wikunia/Javis.jl",
+#)
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -6,11 +6,11 @@ makedocs(;
     authors="Ole Kröger <o.kroeger@opensourc.es> and contributors",
     repo="https://github.com/Wikunia/Javis.jl/blob/{commit}{path}#L{line}",
     sitename="Javis.jl",
-    #format=Documenter.HTML(;
-    #    prettyurls=get(ENV, "CI", "false") == "true",
-    #    canonical="https://Wikunia.github.io/Javis.jl",
-    #    assets=String[],
-    #),
+    format=Documenter.HTML(;
+        prettyurls=get(ENV, "CI", "false") == "true",
+        canonical="https://Wikunia.github.io/Javis.jl",
+        assets=String[],
+    ),
     pages=[
         "Home" => "index.md",
         "References" => "references.md",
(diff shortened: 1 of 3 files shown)


docs/make.jl
```diff
@@ -6,17 +6,18 @@ makedocs(;
...
-    format=Documenter.HTML(;
-        prettyurls=get(ENV, "CI", "false") == "true",
-        canonical="https://Wikunia.github.io/Javis.jl",
-        assets=String[],
-    ),
+    #format=Documenter.HTML(;
+    #    prettyurls=get(ENV, "CI", "false") == "true",
+    #    canonical="https://Wikunia.github.io/Javis.jl",
+    #    assets=String[],
+    #),
```
Wikunia, line 13: Can you clarify why you "deleted" this?

docs/make.jl
```diff
@@ -6,17 +6,18 @@ makedocs(;
...
+        "Contributing" => "contributing.md",
     ],
 )
 
-deploydocs(;
-    repo="github.com/Wikunia/Javis.jl",
-)
+#deploydocs(;
+#    repo="github.com/Wikunia/Javis.jl",
+#)
```
Wikunia, line 23: This is needed to deploy the docs via GitHub actions and should not be "deleted". It probably throws a warning locally but that's normal.

docs/src/contributing.md
```diff
@@ -0,0 +1,67 @@
+# Contributing to Javis
+
```
Wikunia, line 2: Having this file as contributing.md is helpful such that GitHub refers to this document when someone opens a PR. I'm not sure whether it's helpful to have it as part of the documentation.

docs/src/index.md
```diff
@@ -1,6 +1,45 @@
 # Javis
 
-Javis.jl (Julia mAthematical VISualization) is a visualization and animation package that builds on top of [Luxor.jl](https://github.com/JuliaGraphics/Luxor.jl)
+[![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://Wikunia.github.io/Javis.jl/stable)
```
Wikunia, line 3: This file is intended to provide a different introduction than a Readme does. For example more examples and a tutorial for first steps.

docs/make.jl
sudomaze: I have modified make.jl to run my local build without warnings. I will update the code.

docs/make.jl
sudomaze: I see. I thought the CI was causing the issue as when I looked into it online, they mentioned CI caused the error.

``
┌ Warning: Documenter could not auto-detect the building environment Skipping deployment.
└ @ Documenter ~/.julia/packages/Documenter/bRylW/src/deployconfig.jl:75
``

docs/src/contributing.md
sudomaze: @TheCedarPrince suggested including it with the README.md. I think having it would be helpful but it won't be harmful to have as part of the documentation page.

docs/src/index.md
sudomaze: That makes sense. I will reset it to what you have had before. Do you think I can add anything to this page?

docs/src/contributing.md
Wikunia: Oh I see. I thought you moved it but you copied it over, right? I think it makes sense we just need to be careful that we change both if we update the guidelines. 
Or maybe we can just use the contributing.md and build it via Documenter. Not sure whether or needs to be inside the `docs/src/` folder for that.

docs/src/index.md
Wikunia: It should probably include some examples so you could try to add an example with LaTeX bit currently all actual functionality is inside branches so later is probably more interesting.

docs/make.jl
Wikunia: Thanks you should still be able to build it locally. Please comment if that doesn't work. It might throw a warning that it doesn't know where to deploy but the build folder should be correct.

docs/src/index.md
Wikunia: As #6 got merged there are some new features we want to have in a tutorial. Interested @sudomaze

sudomaze: Yeah, sure. I will finish #21 then I will be down to work on adding the examples to the documentation.

Wikunia:
Are you still interested in this? Otherwise I'll close this in 3 days

PR #20, opened 2020-07-28
Created documentation for latex function in compliance with BlueStyle

TheCedarPrince:
I am going to go ahead and merge this as I only added documentation for the `latex` function and cleaned up its syntax. Also, ran tests on it and everything based so in full compliance with the rest of the codebase.

### DIFF
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -11,21 +11,51 @@ const LaTeXSVG = Dict{LaTeXString, String}(
 
 include("svg2luxor.jl")
 
-latex(t::LaTeXString) = latex(t, 10, :stroke)
-latex(t::LaTeXString, font_size::Real) = latex(t, font_size, :stroke)
-latex(t::LaTeXString, action::Symbol) = latex(t, 10, action)
+latex(text::LaTeXString) = latex(text, 10, :stroke)
+latex(text::LaTeXString, font_size::Real) = latex(text, font_size, :stroke)
+latex(text::LaTeXString, action::Symbol) = latex(text, 10, action)
 
 """
-    latex(t::LaTeXString, font_size::Real, action::Symbol)
 
-Add the latex string `text` at the current position. 
-The current position is the top left corner of the latex path. This only works if `tex2svg` is installed.
-It can be installed via:
+`latex(text::LaTeXString, font_size::Real, action::Symbol)`
+
+Add the latex string `text` to the top left corner of the LaTeX path. Can be added to `Luxor.jl` graphics such as `Video` or `Drawing`.
+
+**NOTE: This only works if `tex2svg` is installed.**
+**It can be installed using the following command (you may have to prefix this command with `sudo` depending on your installation):**
+
+> `npm install -g mathjax-node-cli`
+
+# Arguments
+- `text::LaTeXString`: a LaTeX string to render.
+- `font_size::Real`: integer font size of LaTeX string. Default `10`.
+- `action::Symbol`: graphics actions defined by `Luxor.jl`. Default `:stroke`. Available actions:
+  - `:fill` - See `Luxor.fillpath`.
+  - `:stroke` - See `Luxor.strokepath`.
+  - `:clip` - See `Luxor.clip`.
+  - `:fillstroke` - See `Luxor.fillstroke`.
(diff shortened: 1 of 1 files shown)


src/Javis.jl
```diff
@@ -11,21 +11,51 @@ const LaTeXSVG = Dict{LaTeXString, String}(
...
+**NOTE: This only works if `tex2svg` is installed.**
+**It can be installed using the following command (you may have to prefix this command with `sudo` depending on your installation):**
+
+> `npm install -g mathjax-node-cli`
+
+# Arguments
+- `text::LaTeXString`: a LaTeX string to render.
+- `font_size::Real`: integer font size of LaTeX string. Default `10`.
+- `action::Symbol`: graphics actions defined by `Luxor.jl`. Default `:stroke`. Available actions:
+  - `:fill` - See `Luxor.fillpath`.
```
Wikunia, line 33: These actions are not all defined for the command and should be explained in a sentence instead of forcing the reader to look for the documentation somewhere else. 
For this command only `:stroke` is defined `:fill` does the same and is not reasonable for text. Path is defined as well but all others are not.

TheCedarPrince:
Will do! Got a bit overzealous and excited with how the documentation looks. :sweat_smile: Would you like me to revert `master` and ping you again on this branch for a review @Wikunia ?

src/Javis.jl
```diff
@@ -11,21 +11,51 @@ const LaTeXSVG = Dict{LaTeXString, String}(
...
-npm install -g mathjax-node-cli
+using Luxor
+using Javis
+using LaTeXStrings
+
+my_drawing = Drawing(400, 200, "test.png")
+background("white")
+sethue("black")
+latex(L"\\sum \\phi", 100)
+finish()
```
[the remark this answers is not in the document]
TheCedarPrince: You are correct! It should work with only one `\` in front of the LaTeX command. However, when I was trying to render the documentation via:

`help? Javis.latex`

It failed and I had to escape the `\`. Not sure why, but it does work with no issues when I use `\\`.

src/Javis.jl
TheCedarPrince: The reason why I was suggesting that the user refer to the Luxor documentation is that each one of these `Symbol` objects has their own docstrings in Luxor. As Luxor is a different package and the docstrings for each `Symbol` may change at any time, rather than copy the explanations, I just figured I would reference to those docstrings.

I guess I am confused by what you mean that it should be explained in one sentence. Do you mean a sentence like this? This is taken from [Luxor documentation](https://github.com/JuliaGraphics/Luxor.jl/blob/87c3afd565818d847ef10b99a13ab79d80c1a6aa/docs/src/basics.md):

> The default origin is at the top left of the drawing area, but you can reposition it at any time. Many of the drawing functions have an *action* argument. This can be `:none`, `:fill`, `:stroke`, `:fillstroke`, `:fillpreserve`, `:strokepreserve`, `:clip`, or `:path`. The default is `:none`.

Ah, I didn't know that `:stroke` and `:fill` were the only ones defined. So should I just remove the other references to the other `Symbol` objects?

Wikunia:
No need to revert master :) it's more a general advise.

src/Javis.jl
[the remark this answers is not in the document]
Wikunia: Ah okay that's good to know. Good that you tested it. Then leave it as it is.
People who know LaTeXStrings will figure it out 😄

src/Javis.jl
Wikunia: I only support fill and stroke which basically both have the outcome of stroke (but actually filling the path to have thick chars). I also support path because it's useful for animation to not draw it directly but the others are not supported as I don't think that they make sense. (Maybe I'm wrong) in general I would document only stroke and path.
Maybe you can check whether it's possible to link to the documentation of Luxor directly.

TheCedarPrince: That makes sense to me. I'll look into what I can do to link the documentation

Wikunia: Do you mind to create a new PR for this? Where only `path` and `stroke` are explained?

TheCedarPrince: Absolutely! I will open it up now.

Issue #24, opened 2020-08-01
Have a non layered action


Wikunia:
Currently we have the problem that the standard font color is black on a white background.
We can change the color in our `ground` (our normal name for the background function which is called as the first action) but it gets into its own layer such that the color is actually not "leaked" to the other layers.
Unfortunately this is normally what the user expects. It's quite easy to have a keyword argument for this to say:

```
Action(tframes, ground; layered=false)
```

(maybe there is a better name for it @TheCedarPrince ? )
and then everything inside that action is available in the outside world. This way it's also possible to make a shift of the coordinate system there if one doesn't like to have the origin being in the middle.

TheCedarPrince:
Perhaps a better name for this might be "Extend Background Features" as it seems like the background should possibly be a special Action of its own with its own set of features. I think it makes sense to have that layered kwarg as a generic functionality for perhaps any Action. However, for the background Action, with how we are thinking about it, we might want to make it its own unique sort of Action. What do you think? It seems like there may be two problems here that are getting mixed together and by breaking them up, we might be able to better tackle them.

Wikunia:
Okay so we might want to have `BackgroundAction` which has this set to true and we have a kwarg `extend_background`. I think your name is really too long :smile:

TheCedarPrince:
:joy: I tried. How about "Make BackgroundAction"?

TheCedarPrince:
And yea, I agree. That way, we can do unique things with `BackgroundAction` without it affecting other actions.

Wikunia:
How about `in_global_scope` or `in_global_layer`. I like the use of layer as this is what it's actually referring to and people who used Gimp or Photoshop or some others should be able to grasp it easily. Just my thoughts though

TheCedarPrince:
I think I like `in_global_layer` the best because it is more specific and makes me immediately think of the frame layers as opposed to `in_global_scope` as that conveys more than just layers. Let's go with that kwarg.

Wikunia:
Okay that one is merged to I'll close this.

PR #26, opened 2020-08-01
first version for BackgroundAction

Wikunia:
Implements #24 

Currently `BackgroundAction` is not a type but it can easily be in the future. Wherever `Action` is used it should be `ActionType` in the future.


Wikunia:
Do you mind to create a proper GitHub review to show me where I missed docstrings?

Issue #17, opened 2020-07-26
Standardize Documentation Practices


TheCedarPrince:
## Problem 

The documentation for Javis is sparse right now and is not terribly clear in most places. Let's fix that up! I'll be implementing [BlueStyle](https://github.com/invenia/BlueStyle) in most of the areas as best is applicable.

## Places to Improve Documentation

- [ ] **LaTeXSVG** - rather confused about what this is doing (ping @Wikunia for explanation)
- [ ] **float_attribute** - needs a further explanation and definition of kwargs
- [ ] **path_move** - needs a further explanation and definition of kwargs
- [ ] **set_attr** - needs an example
- [ ] **draw_obj** - needs a further explanation and definition of kwargs
- [ ] **include** - rather confused about what this is doing (ping @Wikunia for explanation)
- [ ] **path_quadratic** - needs a further explanation and definition of kwargs
- [ ] **set_attrs** - example and what is the object `o`
- [ ] **eval** - needs a further explanation and definition of kwargs
- [x] **latex** - almost done, needs and example and explanation of actions
- [ ] **pathsvg** - almost done, just needs an example
- [ ] **set_transform** - rather confused about what this is doing (ping @Wikunia for explanation)

Wikunia:
Thanks for the comment..as we have several PRs and this issue now about the open PR #6, we might want to try to merge that then other PRs are easier to review. Maybe you are able to add some missing documentation to it or I'll next week :)

Action is a struct that is holding information about the frames, the function it's calling as well as some transformation.

TheCedarPrince:
Great point @Wikunia !

I am actually going to change this issue to a comprehensive issue on improving general documentation practices in the whole package ecosystem. I'll be working on improving all the documentation as best I can. If I have any questions about anything you are doing, I will ping you!

Wikunia:
From my point of view we probably want to focus on features and documentation of functions which are exported. Nevertheless let me explain the parts you pinged me on here:

LaTeXSVG:
The general way of rendering LaTeX is a two way process. 
1. Render it with a LaTeX compiler into SVG
2. Take the svg and create the drawing
The first step is normally much slower and needs an extra program which I happened to fail to include in the testing pipeline. 
For this reason the first part is cached such that when the user wants to display an equation in several frames the first step is performed once and is then cached. In the frames afterwards the cached svg is simply drawn. 

Can you clarify what you don't understand about `include` or which include you're referring to?
[...]

PR #27, opened 2020-08-02
General documentation

Wikunia:
Might solve some of #17 

1. Added some docstrings to public types and functions
2. The reference in the docs is now split to public and private


Wikunia:
Thanks for the review. I've added a new line before `# Keywords` not sure if needed but is definitely easier to read,

PR #28, opened 2020-08-02
Fixing documentation for latex

TheCedarPrince:
Per @Wikunia 

> These actions are not all defined for the command and should be explained in a sentence instead of forcing the reader to look for the documentation somewhere else.

I am working on a good way of including the documentation here from `Luxor` automatically.


src/Javis.jl
```diff
@@ -11,21 +11,45 @@ const LaTeXSVG = Dict{LaTeXString, String}(
...
+Add the latex string `text` to the top left corner of the LaTeX path. Can be added to `Luxor.jl` graphics such as `Video` or `Drawing`.
+
+**NOTE: This only works if `tex2svg` is installed.**
+**It can be installed using the following command (you may have to prefix this command with `sudo` depending on your installation):**
+
+> `npm install -g mathjax-node-cli`
+
+# Arguments
+- `text::LaTeXString`: a LaTeX string to render.
+- `font_size::Real`: integer font size of LaTeX string. Default `10`.
```
Wikunia, line 31: `::Real` but you say `integer` is a bit confusing :smile: 
It actually supports `Real`

src/Javis.jl
```diff
@@ -11,21 +11,45 @@ const LaTeXSVG = Dict{LaTeXString, String}(
...
+- `action::Symbol`: graphics actions defined by `Luxor.jl`. Default `:stroke`. Available actions:
+  - `:stroke` - See `Luxor.strokepath`.
```
Wikunia, line 33: I would go for: 
- `:stroke` Draws the latex string on the canvas. For more info check `Luxor.strokepath`
- `:path` Creates the path of the latex string but does not render it to the canvas. For more info check [link to the Luxor docs]

src/Javis.jl
```diff
@@ -11,21 +11,45 @@ const LaTeXSVG = Dict{LaTeXString, String}(
...
+latex(text::LaTeXString, font_size::Real) = latex(text, font_size, :stroke)
+latex(text::LaTeXString, action::Symbol) = latex(text, 10, action)
 
 """
-    latex(t::LaTeXString, font_size::Real, action::Symbol)
 
-Add the latex string `text` at the current position. 
-The current position is the top left corner of the latex path. This only works if `tex2svg` is installed.
-It can be installed via:
+`latex(text::LaTeXString, font_size::Real, action::Symbol)`
```
Wikunia, line 20: I think standard form is indented and without "`"

src/Javis.jl
TheCedarPrince: Ooops. Good catch.

src/Javis.jl
TheCedarPrince: That's whay I was thinking about doing too. You're too fast of a reviewer! :joy:

src/Javis.jl
TheCedarPrince: Gotcha. I'll double check BlueStyle again.

src/Javis.jl
Wikunia: Haha sorry. I'm trying to give feedback fast such that you don't need to wait.

TheCedarPrince:
I want to say this is ready to go as we fixed the documentation issue with `Luxor.jl` in #70 and the docstring for `latex` looks fine now. Could I go ahead and merge this @Wikunia ?

PR #30, opened 2020-08-03
testing julia 1.5

Wikunia:


TheCedarPrince:
I tested this as well on my Fedora Linux machine and everything worked as expected. Seems like we are good to go for a merge! I will let you merge it in case you want to add anything else.

As an aside, do you think we could also start looking into benchmarking for Javis? Or do you think that would not be applicable for this project?

Issue #36, opened 2020-08-03
return Point from user defined function


Wikunia:
Currently the user may write:
```
function circ(p=O, color=nothing)
    color !== nothing && sethue(color)
    circle(p, 25, :fill)
    return Transformation(p, 0.0)
end
```
but actually might want to return a `Point` and without an `angle`. 

So the user wants to write

```
function circ(p=O, color=nothing)
    color !== nothing && sethue(color)
    circle(p, 25, :fill)
    return p
end
```

I actually fail to think about a usage for returning an `angle` :smile:

PR #37, opened 2020-08-03
able to return a Point

Wikunia:
Implementation of #36


Wikunia:
Hmm maybe this is not ideal. Maybe we actually want to save the Point and not the transformation...

Wikunia:
Can you clarify what you mean? Currently the `Point` is returned and then transformed into the global scope to be able to use it by other packages.

Wikunia:
It was transformed into a `Transformation` but that was confusing so I removed it with the last commit.

Issue #33, opened 2020-08-03
Non animated Transition 


Wikunia:
Sometimes it might be reasonable to write `Translation(Point(x,y))` which does not depend on the frame instead of writing
`Translation(Point(x,y), Point(x,y))` so I think it makes sense to have a constructor which takes in one value and just uses that one value as the starting and the end of the animation such that there is no animation. Hope that makes sense :smile: 

See #32 as an example.

TheCedarPrince:
This makes sense and I definitely see the utility in this. I was actually thinking about something similar to this earlier today. My thought was along the lines about how if a user wants to draw a circle at the beginning of the animation, then the circle drawing would persist throughout the rest of the animation.

Spawning from that thought was this additional one: Do you think there may be occasion to also make a kwarg for `BackgroundAction` such that if a user only wants actions to run at the beginning of the animation (e.g. drawing shapes or designs on the background of the animation), it will only be run then and not interfere with any of the other Actions? Of course, the animation would need to be redrawn on future frames.

I may be explaining this poorly but what I see beginning to emerge as I am starting to experiment with Javis is that it may be convenient to create clearer distinction between Actions and Background Actions. I am just not sure what would be the best way to continue with clarifying this distinction. Unless you think that Background Actions is sufficient.

Wikunia:
I think I don't really understand what you're referring to and I closed this issue as it's implement, You might want to create a new issue for your thoughts :wink: 

It feels a bit like you're confusing `BackgroundAction` with actions that only show at the beginning? I'm confused :smile: 
The Actions itself always in which frames they are called.

Wikunia:
Okay I want to reopen this for a different reason. It might make sense to see 
`Translation(Point(50, 50))` not as a constant translation but as a translation from the origin to `(50, 50)` 
Actually I think this would be used more often. :confused:

TheCedarPrince:
@Wikunia you are correct. I re-examined my thoughts and I was confusing myself with `Actions`. I might open an issue in the future, but for now, I think we're fine.

Regarding reopening the issue, I see what you're saying. I could see getting confused by that personally. Maybe what might be useful is to make some sort of special kwarg like `anchor` that can be loosely defined as when you want an animation to not be moved from its current position. In code, this could still be seen as `Translation(Point(50, 50))` for `anchor=true` or something along those lines. 

[...]

Issue #38, opened 2020-08-04
Improving Grid Line Animations


TheCedarPrince:
## Background

 As we are finishing up the [Grid lines](#18) initial PR, we are already thinking about how to improve it! This is somewhat lower priority currently but will be worked on consistently.

## Additional Information

Here were some thoughts @Wikunia and I were thinking about implementing next:

> - Have a lighter grid (setline(0.5) or something) and bigger lines every 5 or so steps.
> - Have numbers and titles for the axis
> - The thing about extensibility using Val for example (see #18)

I also thought of a couple that I was exploring:

- Arrows at the end of lines
- Defining x and y axis limits for zero-lines

If anyone else has any thoughts, feel free to drop a comment/thought! For reference, here is what the current grid lines animation look like:

![grid_animation](https://user-images.githubusercontent.com/29561456/89257930-23a5e580-d5f5-11ea-8426-7620e8e0096d.gif)

Wikunia:
When you're working on this also make sure to change the test cases slightly to have better coverage.
i.e you currently test 4 animations but all use the same `line_gap` and `line_thickness`. 

It might also be valuable to set the color maybe even two such that every second line is using a different color.

Wikunia:
Maybe we should use SubAction `appear` for this as currently I feel like there is no way to actually have a drawn grid which does not disappear after it's drawn completely.

TheCedarPrince:
That makes sense to me @Wikunia.

Furthermore, the animation here does not actually "mean" anything. What I mean by that is that I can't draw vectors or objects in relation to the grid. I'd like to change that - see #98

TheCedarPrince:
See Ole's comment here in issue [97](https://github.com/Wikunia/Javis.jl/issues/97#issuecomment-675387317).

muhammadalics:
Hi! Is this project still up for grabs?

TheCedarPrince:
Hey @muhammadalics - I have received your email. Let's discuss there. :)

PR #46, opened 2020-08-06
Transition shorthand

Wikunia:
As discussed in #33 we decided to revert `Translation(p)` which mapped to `Translation(p, p)` and now map it to `Translation(O, p)`

Additionally one can now write `Translation(x, y)` such that one does not need to write `Point` every time.

### DIFF
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -128,10 +128,17 @@ end
 """
     Translation(p::Union{Point, Symbol})
 
-Create a `Translation(p, p)` such that 
-if there is no animation the start position and end position are the same.
+Create a `Translation(O, p)` such that a translation is done from the origin.
 """
-Translation(p::Union{Point, Symbol}) = Translation(p, p)
+Translation(p::Union{Point, Symbol}) = Translation(O, p)
+
+"""
+    Translation(x::Real, y::Real)
+
+Create a `Translation(O, Point(x,y))` such that a translation is done from the origin.
+Shorthand for writing `Translation(Point(x,y))`.
+"""
+Translation(x::Real, y::Real) = Translation(Point(x, y))
 
 """
     Rotation <: Transition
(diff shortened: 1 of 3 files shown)


src/Javis.jl
```diff
@@ -128,10 +128,17 @@ end
...
+
+"""
+    Translation(x::Real, y::Real)
+
+Create a `Translation(O, Point(x,y))` such that a translation is done from the origin.
+Shorthand for writing `Translation(Point(x,y))`.
+"""
+Translation(x::Real, y::Real) = Translation(Point(x, y))
 
 """
```
TheCedarPrince, line 143: Since we decided to only thoroughly document exported functions, could you add a line of an example invocation of this function? To show how it can be used ?

src/Javis.jl
```diff
@@ -152,10 +159,10 @@ end
...
     Rotation(r::Union{Float64, Symbol})
 
-Rotation as a transition from `r` to itself.
-Can be used as a general non-animating rotation.
+Rotation as a transition from 0.0 to `r` .
+Can be used as a short-hand.
 """
-Rotation(r::Union{Float64, Symbol}) = Rotation(r, r)
+Rotation(r::Union{Float64, Symbol}) = Rotation(0.0, r)
```
TheCedarPrince, line 166: Since we decided to only thoroughly document exported functions, could you add a line of an example invocation of this function? To show how it can be used ?

Wikunia:
Thanks for the review. Can you clarify on what you mean by your point about Documenter? They should all be normal docstrings like others.

I followed docstring practice as for the currently existing docstrings for Transformation and Rotation. I can have a link to `javis` in all to give an example but I think it's counter productive. I'll write on Slack 😬

I can't think of a unit test for #18 so I think the current tests are ok (mentioned improvements in #38 .

Issue #22, opened 2020-07-31
Easier way of defining frames


Wikunia:
I have the feeling that defining frames for an Action is probably a tedious task for a full video or a more complex animation.
We should think about different ways like:
- `:all` for all frames
- `:same` same as the one before
- `Rel(10:20)` relative to the start of the one before so starting 10 frames after the one before and then going for 11 frames.

TheCedarPrince:
Oh I really like this idea @Wikunia. Especially the `:same` symbol as that could potentially better ease the `FFMPEG` backend by making it not have to process as much.

As of right now, I am actually rather confused by how the frames work when called from an `Action` in the `javis` function. Would be happy to work on some documentation about it if you could explain it here a bit more.

Wikunia:
I think we have a small miscommunication. I mean `:same` as the action before not as the frame before which should not happen that often or can have a special action called `Wait` or something which just uses the frame before. 

I am also thinking whether it's possible to create a video faster than creating images and use ffmpeg the way we do. 

Frames how they work now:
It's simple and stupid maybe 😄
i.e. 1:100 means this action is shown in frames 1 to 100. So for 30fps this means a bit more than 3 seconds.

We might want to define a way to use seconds as well to easily change the frame rate later. A lot can be done to make it easier for the user I think.

Wikunia:
Let's have:
`:same` as the default
i.e
```
 Action(1:100, ground; in_global_layer=true),
 Action(1:90, :red_ball, (args...)->circ(p1, "red"), Rotation(from_rot, to_rot)),
 Action(Rel(10), :blue_ball, (args...)->circ(p2, "blue"), Rotation(2π, from_rot, :red_ball)),
 Action((video, args...)->path!(path_of_red, pos(:red_ball), "red")),
```
is the same as:
```
 Action(1:100, ground; in_global_layer=true),
 Action(1:90, :red_ball, (args...)->circ(p1, "red"), Rotation(from_rot, to_rot)),
 Action(91:100, :blue_ball, (args...)->circ(p2, "blue"), Rotation(2π, from_rot, :red_ball)),
 Action(91:100, (video, args...)->path!(path_of_red, pos(:red_ball), "red")),
```

Issue #25, opened 2020-08-01
Tutorials for v0.1.0 Release


Wikunia:
Maybe this is good to be part of a tutorial on how to use Javis?

![trigo](https://user-images.githubusercontent.com/4931746/89107711-37a1e980-d433-11ea-9f57-7e64d3d49cb7.gif)

Thoughts @TheCedarPrince ?

TheCedarPrince:
Oh that would be wonderful! I like this idea a lot! How do you think we should distribute tutorials? Like a Jupyter notebook or a markdown follow-along sort of tutorial where you follow directions?

Taking inspiration from [divio](https://documentation.divio.com/tutorials/)

Wikunia:
I would do this similar to the one in Luxor.jl as we both found that one quite nice.

TheCedarPrince:
Hey @sudomaze, any interest in taking this part on?

TheCedarPrince:
Some thoughts for tutorials:

1. Making your first gif
2. How to make your own functions for Javis
3. Check out divio again for ideas [divio](https://documentation.divio.com/tutorials/)

TheCedarPrince:
Add a tutorial about what actions are and how to use them (especially with subactions)

TheCedarPrince:
Tutorials that will be worked on for v0.1.0 release of `Javis.jl`:

- [ ] An easy entry to making your first animation with `Javis.jl`
- [ ] What are Actions? How do they work? (include discussion about Subactions and Rel/:same)
- [ ] Javis can render LaTeX?!?!?

PR #47, opened 2020-08-06
support for :same and Rel as frames

Wikunia:
Implements #22

### DIFF
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -1,6 +1,8 @@
 module Javis
 
-using Luxor, LaTeXStrings
+using LaTeXStrings
+using LightXML
+using Luxor
 
 """
     Video
@@ -17,6 +19,7 @@ mutable struct Video
     height  :: Int
     defs    :: Dict{Symbol, Any}
 end
+
 """
     Video(width, height)
 
@@ -33,8 +36,12 @@ function Video(width, height)
     return video
 end
 
-# holds the current video (Array to be declared as constant :D)
-# only holds one video at a time 
+"""
+    CURRENT_VIDEO
+
+holds the current video in an array to be declared as a constant
+The current video can be accessed using CURRENT_VIDEO[1]
+"""
 const CURRENT_VIDEO = Array{Video, 1}()
 
 """
@@ -101,6 +108,9 @@ Action(1:90, :red_ball, (args...)->circ(p1, "red"), Rotation(from_rot, to_rot)),
 Action(91:100, :blue_ball, (args...)->circ(p2, "blue"), Rotation(2π, from_rot, :red_ball)),
 Action(91:100, (video, args...)->path!(path_of_red, pos(:red_ball), "red"))
 ```
+
+# Fields
+- rel::UnitRange defines the frames in a relative fashion. 
 """
 struct Rel
     rel :: UnitRange
@@ -359,9 +369,9 @@ Add the latex string `text` to the top left corner of the LaTeX path. Can be add
 # Example
 
 ```
-using Luxor
 using Javis
 using LaTeXStrings
+using Luxor
 
 my_drawing = Drawing(400, 200, "test.png")
 background("white")
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -21,8 +21,21 @@ end
     Video(width, height)
 
 Create a video with a certain `width` and `height` in pixel.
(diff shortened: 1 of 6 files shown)


src/Javis.jl
```diff
@@ -21,8 +21,21 @@ end
     Video(width, height)
 
 Create a video with a certain `width` and `height` in pixel.
+This also sets `CURRENT_VIDEO`.
```
TheCedarPrince, line 24: I think it would be wise to explain what this variable's purpose is and why we are using it here.

src/Javis.jl
```diff
@@ -69,6 +82,37 @@ mutable struct Action <: ActionType
     opts                    :: Dict{Symbol, Any}
 end
 
+"""
+    Rel
+
+Ability to define frames in a relative fashion.
+
```
TheCedarPrince, line 89: Similar to what you did for the docstring for `Video` could you also define the fields of Rel?

src/Javis.jl
```diff
@@ -79,18 +123,78 @@ The most simple form of an action (if there are no `args`/`kwargs`) just calls
 Action(frames, func::Function, args...; kwargs...) = Action(frames, nothing, func, args...; kwargs...)
 
 """
```
TheCedarPrince, line 125: I think this docstring is good but since we are so heavily extending what an `Action` is, we might want to link to some bigger tutorial here about different invocations of all the different Actions we have made. That way, a user could now exactly what the differences are between each form. (ping #25 )

test/animations.jl
```diff
@@ -59,16 +59,16 @@ end
...
-	    
-	    Action(11:20, draw_grid(direction = "BR", line_gap = 25)),
-	    Action(11:20, zero_lines(direction = "BR", line_thickness = 10)),
-	    
-	    Action(21:30, draw_grid(direction = "TL", line_gap = 25)),
-	    Action(21:30, zero_lines(direction = "TL", line_thickness = 10)),
+	    Action(zero_lines(direction = "BL", line_thickness = 10)),
 	    
-	    Action(31:40, draw_grid(direction = "TR", line_gap = 25)),
-	    Action(31:40, zero_lines(direction = "TR", line_thickness = 10)),
```
TheCedarPrince, line 71: This looks so great now with Rel invoked! :smile:

test/runtests.jl
```diff
@@ -4,6 +4,6 @@ using Images
 using LaTeXStrings
 using Test, ReferenceTests
 
+include("unit.jl")
 include("svg.jl")
 include("animations.jl")
```
TheCedarPrince, line 9: According to BlueStyle, these includes actually should be in alphabetical order.

test/runtests.jl
```diff
@@ -4,6 +4,6 @@ using Images
 using LaTeXStrings
 using Test, ReferenceTests
```
TheCedarPrince, line 5: According to BlueStyle, packages should each be on their own line and each `using` should be in alphabetical order.

test/runtests.jl
Wikunia: I don't think this applies for tests as I want to run unit tests first as that makes sense in my opinion.

TheCedarPrince: I agree with you on principal. However, it seems like this section here on [module imports]( https://github.com/invenia/BlueStyle#module-imports) and [test formatting](https://github.com/invenia/BlueStyle#test-formatting) are at odds somewhat.

Just keep the order you have unless if you think it should change based on BlueStyle are my thoughts.

src/Javis.jl
Wikunia: In documenter they will be all underneath each other such that this is the first which holds no real information as the actual examples are given in the later `Actions` if that makes sense? There `kwargs` and `args` are explained. I also do link to `javis` for usage. I'm confused by what you want to add here I guess.

TheCedarPrince: Oh, it was more just a general remark. Resolving issue.

src/Javis.jl
Wikunia: Done. Good point!

test/runtests.jl
Wikunia: I didn't find specifics on that as they mention `using` where the order is irrelevant but for `include` it matters.

PR #54, opened 2020-08-07
using FFMPEG

### LINKED ISSUE #53: ffmpeg not installed by default on windows
So I can't use javis which is calling ffmpeg.

I had to install it and set my path. 99% of users won't know how to do it.

Perhaps binarybuilder.jl can help?

I want to help but I may not get to it soon enough.

Wikunia:
Fix #53 

This should remove the necessity to install FFMPEG which might be a pain for Windows users.


Wikunia:
Do you mind to test that @xiaodaigh ? Should work as it works in GitHub actions.

Wikunia:
Test by @xiaodaigh worked. Can I merge @TheCedarPrince ?

Issue #57, opened 2020-08-08
Better defaults for kwargs in javis


Wikunia:
Currently the defaults for `tempdirectory`, `deletetemp`, ... are all not the best :smile: 

I think `tempdirectory` should be https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.tempdir or maybe `mktempdir`

- `pathname = javid_randomstring.gif`

I would remove `creategif` and just create a `gif` if pathname is not empty. This makes it easy to later set `pathname` to `filename.mp4` to create a video and possibly we can also remove `deletetemp`. :wink: 

The bug label is coming from this error message:
```
IOError: could not spawn `rm -rf .`: no such file or directory (ENOENT)
```
mentioned in slack when using `deletetemp`

TheCedarPrince:
Really like this idea. I approved it over in #58

PR #58, opened 2020-08-08
Reasonable defaults

Wikunia:
better defaults as described in #57

### DIFF
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -4,6 +4,7 @@ using FFMPEG
 using LaTeXStrings
 using LightXML
 using Luxor
+using Random
 
 """
     Video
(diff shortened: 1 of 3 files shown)


src/Javis.jl
```diff
@@ -647,15 +646,17 @@ function javis(
...
+    isempty(pathname) && return
+    path, ext = splitext(pathname)
+    if ext == ".gif"
+        # generate a colorpalette first so ffmpeg does not have to guess it
+        ffmpeg_exe(`-loglevel panic -i $(tempdirectory)/%10d.png -vf "palettegen=stats_mode=diff" -y "$(tempdirectory)/palette.bmp"`)
+        # then apply the palette to get better results
+        ffmpeg_exe(`-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png -i "$(tempdirectory)/palette.bmp" -lavfi "paletteuse=dither=sierra2_4a" -y $pathname`)
+    else
+        @error "Currently it's only supported to create a gif and not a $ext."
+    end
```
TheCedarPrince, line 658: English grammar mistake here. It should read:

> Currently, only gif creation is supported and not a $ext.

Wikunia: Oh yeah and I should add a test case for that...

PR #65, opened 2020-08-09
Bugfix for short-hand rotation

### LINKED ISSUE #60: Rotation from 0 by default
I somehow missed this in one of my PRs :confused: 

See: https://github.com/Wikunia/Javis.jl/blob/master/src/Javis.jl#L274-L288

Wikunia:
Fix #60


Wikunia:
We both missed it. I will also from now on describe my PRs better to make them easier to review. New one is coming in a few minutes :wink:

Issue #62, opened 2020-08-08
Access to arbitrary return


Wikunia:
If a drawing function returns something other than a `Point` or a `Transformation` the user currently can't access it easily.
We might want to have a function `val` for that.
Such that 
```
function ret(args...)
   return "Bla"
end
```

```
Action(1:100, :test, ret)
Action(1:100, (args...)->text(val(:test))
```

works

TheCedarPrince:
I really like this idea. Let me know if you want me to test this out at some point when ready. I will add it into the tutorials when it is ready to go. #25

PR #67, opened 2020-08-10
Introduction of `val` and added some docstrings for pos, ang, val

Wikunia:
Ability to use arbitrary returns as explained in #62 

Example:
```
javis(video, [
      BackgroundAction(1:50, :framenumber, ground),
      Action(:same, (args...)->text(val(:framenumber), Point(0, -200))),
], pathname="test.gif")


function ground(video, action, framenumber) 
    background("white")
    sethue("blue")
    return "Frame: $framenumber"
end
```


Wikunia:
I assume the current problem is having a different font for different systems. 
Using LaTeX will probably fix this

src/Javis.jl
```diff
@@ -491,28 +500,49 @@ function perform_transformation(trans::InternalRotation)
     rotate(trans.angle)
 end
 
-pos(p::Point) = p
-pos(t::Transformation) = t.p
+"""
+    val(s::Symbol)
+
+Get access to the value that got saved in `s` by a previous action.
+If you want to access a position or angle check out [`pos`](@ref) and [`angle`](@ref).
```
TheCedarPrince, line 507: Similarly to what we discussed with `ang` and `pos` in #66, can you also give an alternative name like `get_value`? Thanks!

src/Javis.jl
```diff
@@ -491,28 +500,49 @@ function perform_transformation(trans::InternalRotation)
...
-end
+"""
+    angle(s::Symbol)
+
+Get access to the angle that got saved in `s` by a previous action.
+
+# Returns
+- `Float64`: the angle stored by a previous action i.e via `return Transformation(p, angle)`
+"""
+angle(s::Symbol) = angle(val(s))
```
TheCedarPrince, line 545: Merge conflict with #66 based on naming -- just a heads-up for you to be aware of.

src/Javis.jl
```diff
@@ -695,7 +725,7 @@ end
 export javis, latex
 export Video, Action, BackgroundAction, Rel
 export Line, Translation, Rotation, Transformation
-export pos, angle
+export val, pos, angle
```
TheCedarPrince, line 728: If you add the alternative name for `val`, don't forget to add it here.

test/animations.jl
```diff
@@ -110,13 +110,19 @@ end
     end
 end
 
+function disp_text(t, p)
+    setfont("Helvetica", 24)
+    settext(t, p)
+end
+
```
TheCedarPrince, line 117: Error here due to not all systems possibly possessing this font. Not sure how to get around this.

src/Javis.jl
Wikunia: I'll do this in #66 and merge that one first then.

src/Javis.jl
```diff
@@ -491,28 +500,49 @@ function perform_transformation(trans::InternalRotation)
...
+
+"""
+    pos(s::Symbol)
+
+Get access to the value that got saved in `s` by a previous action.
+If you want to access a position or angle check out [`pos`](@ref) and [`angle`](@ref).
+
+# Returns
+- `Point`: the point stored by a previous action.
+"""
```
Wikunia, line 532: stupid wrong copy paste mistake

test/animations.jl
Wikunia: I just changed it to setting the radius of a circle instead.

src/Javis.jl
Wikunia: Yeah it was a hard one but I think I managed it.

Issue #70, opened 2020-08-11
Auto-Export of Luxor Functions


TheCedarPrince:
Based on conversation with @Wikunia , the manual importing of `Luxor` is somewhat of a pain for each file written using Javis. @thofma suggested a great idea based on [Hecke.jl](https://github.com/thofma/Hecke.jl/blob/da91ba6cd304e4399a272931b022d5d94cb407b2/src/Hecke.jl#L82) to auto-export each function in Luxor automatically within Javis. Let's get this implemented for documentation and for #44 .

thofma:
If you are not afraid of additional dependencies, you could also use the Reexport.jl package like here: https://github.com/SciML/DifferentialEquations.jl/blob/master/src/DifferentialEquations.jl

Wikunia:
Yeah we kinda thought it's not necessary as you mentioned and we have more control over some behavior of exporting only some functions and overwriting others I guess.

PR #71, opened 2020-08-11
Auto export of Luxor functions

TheCedarPrince:
Implements #70 -- let me know if there is anything else you want to tweak from this!


Wikunia:
You should remove `using Luxor` from the test cases.

Issue #43, opened 2020-08-04
Let a shape, text, whatever appear


Wikunia:
I think it would be nice to let something be drawn like a polygon line by line or let it appear from the background by blending in.
Blending in can be change in opacity or line thickness.

Wikunia:
It feels like we need #44 for that @TheCedarPrince ?
Or how else shall the user define appear and disappear?
```
Action(1:100, func, Appear(10), Disappear(10))
```
which means Appear in the first 10 frames then stay and disappear in the last ten might work but with subactions this will be much cleaner.

TheCedarPrince:
I agree - I think we may want to wait until we have subactions functioning and then implement this fully. I think that without subactions, this sort of call looks strange and not intuitive. Now that we also have #47 implemented, I am excited because I think things are going to start looking much cleaner when we also implement subactions.

Wikunia:
Shall I change the milestone of #44 to v0.1.0 or this one to v0.2.0? If I change this one basically only tutorial is missing from v0.1.0 so I tend to implement SubActions earlier ;)

TheCedarPrince:
Yea, change it to v0.1.0.

PR #72, opened 2020-08-11
Subactions

Wikunia:
This is a first version of subactions:

nicholas

can be drawn with:

```
using Javis

function ground(args...) 
    background("white")
    sethue("black")
end

function house_of_nicholas(p1=O, width=100, color="black")
    # sethue(color)
    #     .p1
    # .p2   .p3
    #
    # .p4   .p5
    width2 = div(width, 2)
    p2 = p1 + Point(-width2, width2)
    p3 = p1 + Point(width2, width2)
    p4 = p2 + Point(0, width)
    p5 = p3 + Point(0, width)
    setline(3)
    line(p4, p5, :stroke)
    line(p5, p2, :stroke)
    line(p2, p4, :stroke)
    line(p4, p3, :stroke)
    line(p3, p1, :stroke)
    line(p1, p2, :stroke)
    line(p2, p3, :stroke)
    setline(8)
    line(p3, p5, :stroke)
end

function anim_house_of_nicholas()
    demo = Video(500, 500)
    javis(demo, [
        BackgroundAction(1:150, ground),
        Action((args...)->house_of_nicholas(); subactions = [
            SubAction(1:20, appear(:fade)),
            SubAction(81:100, disappear(:fade))
        ])
    ], pathname="current/nicholas.gif")
end
```

**Important** the refs changed because I think the default line width should be the same as `setline(1)` as that feels the most natural. What do you think @TheCedarPrince ?

Some current things that are missing:
- [x] If someone sets `setline` in a BackgroundAction the default of subactions should be updated.
- [x] Possibility to do the same with opacity and set that as default as for `:fill` the current one is stupid :smile: 
- I'm not sure when to move over to #43 :wink: @TheCedarPrince


Wikunia:
Another example of fading in with opacity:
circle_square

src/Javis.jl
```diff
@@ -69,6 +72,52 @@ abstract type InternalTransition end
...
 abstract type ActionType end
 
+"""
+    SubAction <: ActionType
+
+Defines a subaction with relative frames and a function.
+
+# Fields
+- `frames::UnitRange{Int}`: the frames relative to the parent [`Action`](@ref)
+- `func::Function`: the function that gets called in each of those frames. Takes the following arguments: video, action, subaction, rel_frame
```
TheCedarPrince, line 82: Could you make sure this line is under or just at 92 characters? https://github.com/invenia/BlueStyle#vs-code-settings

src/Javis.jl
```diff
@@ -770,9 +885,13 @@ for func in names(Luxor; imported=true)
...
 
 export javis, latex
-export Video, Action, BackgroundAction, Rel
+export Video, Action, BackgroundAction, SubAction, Rel
 export Line, Translation, Rotation, Transformation
 export val, pos, ang, get_value, get_position, get_angle
 export projection, morph
+export appear, disappear
+
+# luxor extensions
```
TheCedarPrince, line 894: Suggestion:

# Luxor extensions -> # Custom Override of Luxor Functions

src/luxor_extensions.jl
```diff
@@ -0,0 +1,31 @@
+"""
+    setline(linewidth)
+
+Set the line width and multiplies it with the current multiplier which is i.e. set by [`appear`](@ref) and [`disappear`](@ref).
+
```
TheCedarPrince, line 5: Since these files are exported, could you add Example invocations?

src/subaction_animations.jl
```diff
@@ -0,0 +1,67 @@
+"""
+    appear(s::Symbol)
+
+Appear can be used inside a [`SubAction`](@ref) i.e.
+```
+Action(101:200, (args...)->house_of_nicholas(); subactions = [
+            SubAction(1:20, appear(:fade)),
+            SubAction(81:100, disappear(:fade))
+        ])
+```
```
TheCedarPrince, line 10: Could you add the example tag? Thanks! :smile:

src/subaction_animations.jl
```diff
@@ -0,0 +1,67 @@
...
+"""
+    disappear(s::Symbol)
+
+Disappear can be used inside a [`SubAction`](@ref) i.e.
+```
+Action(101:200, (args...)->house_of_nicholas(); subactions = [
+            SubAction(1:20, appear(:fade)),
+            SubAction(81:100, disappear(:fade))
+        ])
+```
```
TheCedarPrince, line 44: Could you add the example tag? Thanks! :smile:

Wikunia:
This currently does not support any transformations as they are structs and not callable functions. I have to think about that for a bit but I think it should be implemented before merging this

Wikunia:
@TheCedarPrince what do you think about if

```julia
function circ(p=O, color="black")
    sethue(color)
    circle(p, 25, :fill)
    return p
end

video = Video(500, 500)
javis(video, [
        Action(1:70, ground; in_global_layer=true),
        Action(:same, :red_ball, (args...)->circ(O, "red"), Rotation(0, 2π)),
])
```

can also be written as:
```julia
function circ(p=O, color="black")
    sethue(color)
    rotate(2π)
    circle(p, 25, :fill)
    return p
end

video = Video(500, 500)
javis(video, [
        Action(1:70, ground; in_global_layer=true),
        Action(:same, :red_ball, (args...)->circ(O, "red"); subactions=[
            SubAction(:same, animate_transition())
       ])
])
```

Wikunia:
Okay I think it's good to provide the extra in v0.2.0 and not now but how about:

```
function circ(p=O, color="black")
    sethue(color)
    circle(p, 25, :fill)
    return p
end

video = Video(500, 500)
javis(video, [
        Action(1:70, ground; in_global_layer=true),
        Action(:same, :red_ball, (args...)->circ(O, "red"); subactions=[
              SubAction(1:70, Rotation(0, 2π))
        ]
])
```

which also does not work currently but I think it should work in v0.1.0.

Wikunia:
I've built in support for transitions inside subactions:

```julia
function ground_opacity(args...)
    background("white")
    sethue("darkblue")
    setopacity(1.0)
end

function circ(p=O, radius=25, color="black")
    sethue(color)
    circle(p, radius, :fill)
end

function circle_and_square()
    demo = Video(500, 500)
    javis(demo, [
        BackgroundAction(1:200, ground_opacity),
        Action((args...)->circ(O, 50, "darkblue"); subactions = [
            SubAction(1:25, appear(:fade)),
            SubAction(26:75, Translation(Point(100, 20))),
            SubAction(76:100, Translation(Point(20, -40))),
            SubAction(101:175, Translation(Point(-120, 20))),
            SubAction(176:200, disappear(:fade))
        ]),
    ], tempdirectory="current/images", pathname="current/circle_square.gif")
end
```

produces

circle_square

I decided that it makes sense for transitions to be persistent which means after frame 75 this
```
SubAction(26:75, Translation(Point(100, 20))),
```
still gets called so it translates to Point(100, 20)) even for framenumber `> 75`

This is not the case for function so if use 300 frames for the action and still use
```
SubAction(176:200, disappear(:fade))
```
the circle will be there at frame 201 instead of staying disappeared. Does that make sense @TheCedarPrince or shall I change that behavior?

Wikunia:
I've implemented now that the subaction is called with the last frame such that disappeared objects stay disappeared. I think that makes the most sense.

src/Javis.jl
```diff
@@ -67,32 +71,158 @@ end
...
+A `SubAction` can be defined with frames and a function
+inside the `subactions` kwarg of an [`Action`](@ref).
+In the following example the `house_of_nicholas` (which is not part of Javis)
+appears in the first 20 frames which means the opacity is increased from 0 to 1.0.
+Then it stays at full opacity and disappears the same way in the last 20 frames.
+
+# Example
+javis(demo, [
+    BackgroundAction(1:100, ground),
+    Action((args...)->house_of_nicholas(); subactions = [
```
TheCedarPrince, line 105: I think that instead of using the `house_of_nicholas` function, instead, just use the `circ` func we generally use. The reason for this is the `circ` func is small enough to include here and also a straightforward example. If you had not explained the other day what the `house_of_nicholas` was, I personally would have gotten confused by this example and what you were trying to show.

It's a great idea, just maybe not use house of nicholas here. Perhaps in a tutorial or something?

src/Javis.jl
```diff
@@ -67,32 +71,158 @@ end
...
 abstract type InternalTransition end
 
-abstract type ActionType end
+abstract type AbstractAction end
+
+"""
+    SubAction <: AbstractAction
+
+Defines a subaction with relative frames and a function.
+
```
TheCedarPrince, line 80: I see that you explain how the subaction works in the following function definitions but it is actually not explicitly stated "what" a subaction is and why you'd want to use them. Could you add a bit of exposition about what a SubAction actually is? Thanks!

Wikunia: I hope it's clearer now. In general the example in the constructor should be quite simple to understand, I hope.

src/Javis.jl
Wikunia: Good point!

Wikunia:
I just tried to merge master via GitHub so I hope I didn't mess things up. Let's see whether the tests pass :smile:

PR #82, opened 2020-08-13
call subaction with rel_frame and test case

Wikunia:
The subaction transformation was called with `frame` but it should be `rel_frame`.


Wikunia:
You can thank Jarvis. He found the bug

Issue #78, opened 2020-08-13
Tutorial about Transitions and SubActions


Wikunia:
One big tutorial an all those things. I try to make something fun and maybe use the result of Tutorial one for this.

Wikunia:
I'm sorry but that's what I currently have:

![tutorial_3](https://user-images.githubusercontent.com/4931746/90175569-e89b7300-dda7-11ea-9469-ad9fb1684ad8.gif)

PR #84, opened 2020-08-13
Tutorial #4: The creation of Jarvis

Wikunia:
See #78


docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
+# **Tutorial 4:** Do you know our Mascot? - Learn about transitions and subactions!
+
+You have learned a couple of cool features of Javis already. Now you're ready to finally meet our little mascot. Well actually you can't just see him, we have to create him first. :smile:
```
TheCedarPrince, line 3: You'll need to manually add in the emoji. I generally use the julia repl and do `clipboard("\:smile:")` and that works.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
+# **Tutorial 4:** Do you know our Mascot? - Learn about transitions and subactions!
```
TheCedarPrince, line 1: Can you convert this to titlecase? Thanks!

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+
+You have learned a couple of cool features of Javis already. Now you're ready to finally meet our little mascot. Well actually you can't just see him, we have to create him first. :smile:
+
```
TheCedarPrince, line 4: Could you add a section here about learning outcomes? Meaning, what a user will learn by following this tutorial?

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+Let's create a list of what we want first
+- a circular head
+- some hair
+- eyes
+- a nose
+- a moving mouth
+- he should say something
```
TheCedarPrince, line 13: Can you make this the start of its own section?

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+
+This should show you the power of `SubAction` so we want each of those appear (fade in) one after another.
```
TheCedarPrince: I'd suggest putting this line after you define the title and background so you can better and more clearly introduce a SubAction.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+    javis(video, [
+        BackgroundAction(1:150, ground),
+        Action(title; subactions=[
+            SubAction(1:5, appear(:fade)),
+        ]),
+    ], pathname="jarvis.gif", framerate=15)
+end
+```
+
+The `ground` function should be familiar to you as well as the general structure of the `javis` function.
```
TheCedarPrince, line 46: Put this before the code block and explain how this should be common now.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+
+A new part is the second `Action` which takes in a function and `subactions` as a keyword argument. 
+
+If you're wondering where the `frames` are specified:
+- You can leave the frames blank as here and then the frames from the previous action are used. 
+
+The `subactions` keyword uses a list of [`SubAction`](@ref) structs which are defined in a similar fashion with `frames` and a function but are in some sense simpler than the `Action`.
```
TheCedarPrince, line 53: Could you explain here why we are using SubActions and why someone would want to use a SubAction?

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+
+```julia
+Action(16:150, (args...)->circle(O, 100, :stroke); subactions=[
+    SubAction(1:15, appear(:fade)),
+]),
+```
+
+This is very similar to the previous action. Here we can see that `SubAction` uses relative frame numbers such that the head appears in the frames `16:30` and then is at full opacity afterwards.
+
+Just a small refresher: We need the anonymous function `(args...)->circle(O, 100, :stroke)` as each function gets called with the three arguments `video, action, frame`.
```
TheCedarPrince, line 73: I'd suggest making this a Note block like this:

```mkd
> **NOTE:** blah blah
```

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+
+Okay let's add some hair shall we?
```
TheCedarPrince, line 75: I'd suggest starting a new section here.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,304 @@
...
+We can use splatting for that :wink:
+
+```julia
+[
+    Action(26:150, (args...)->hair_blob(hair_angle[i]); subactions=[
+        SubAction(1:25, appear(:fade)),
+    ]) for i=1:20
+]...,
+```
+
```
TheCedarPrince, line 108: I really like this! :smile: However, you may want to explain what you are doing here exactly with splatting in this function.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,310 @@
...
+])
+```
+
+This time we also use the [`disappear`](@ref) function to fade out the text.
+
+## Final outcome and code
+
+![Jarvis](./assets/jarvis.gif)
+
+Hope you had as much fun reading this tutorial as I had creating our mascot.
```
TheCedarPrince, line 216: Could you summarize the learning outcomes here stated at the beginning? Thanks!

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,310 @@
...
+    do_action(:stroke)
+end
+```
+
+This function uses some more functions of the awesome Luxor package.
+
+The lips should be a little thicker than the other lines that we have drawn so far so let's set `setline(2)`. (default is 1).
+First we move to the starting point of the lip and create two control points a bit below and to the vertical center.
+
+The `curve` function is used to draw a bezier curve.
```
TheCedarPrince, line 167: I'd suggest maybe a small note on a bezier curve that links to the wikipedia article on bezier curves if anyone has questions.

docs/src/tutorials/tutorial_4.md
Wikunia: I think I explained it now in a few different ways and they are not that different from `Action`. If you think you don't understand how they work it would be great if you can clarify what you didn't understand or what users might not understand yet.

docs/src/tutorials/tutorial_4.md
Wikunia: I changed this to the learning outcome now so have it still at the top.

docs/src/tutorials/tutorial_4.md
Wikunia: thanks

docs/src/tutorials/tutorial_4.md
Wikunia: Hope it's more clear now

docs/src/tutorials/tutorial_4.md
Wikunia: Good point.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,329 @@
...
+            SubAction(1:5, appear(:fade)),
+        ]),
+    ], pathname="jarvis.gif", framerate=15)
+end
+```
+
+A new part is the second `Action` which takes in a function and `subactions` as a keyword argument this time.
+
+If you're wondering where the `frames` are specified:
+- You can leave the frames blank as here and then the frames from the previous action are used.
```
TheCedarPrince, line 51: I would suggest this phrasing:

```mkd
A new component introduced in this tutorial is shown in the second `Action`.
It is the kwarg `subactions` followed by the object `SubAction`.
A `SubAction` is invoked when you want to apply an additional action or provide an additional functionality to the original `Action`.
Although one could use standard an `Action` to achieve the same functionality via an `id`, `subactions` provide a much easier and more organized method that also improves readability.

With the invocation of `SubAction` in the `face` function, the function can be broken down as follows:
- We define our set the background for the animation using a `BackgroundAction`.
- Using the `title` function, we invoke an `Action` and then provide a `SubAction.`
  - The `SubAction` utilizes the `Javis.jl` provided function, `appear` to cause the title to slowly appear over the `UnitRange` of `1:5`. 

> **NOTE:** For a `SubAction` you can leave the frames arg blank. 
> The frames from the previous action are used. 
```

docs/src/tutorials/tutorial_4.md
TheCedarPrince: I think this explanation is good now coupled with my suggestion in the prior conversation.

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,329 @@
...
+- eyes
+- a nose
+- a moving mouth
+- he should say something
+
+## Learning Outcomes
+
+This tutorial should show you the power of subactions which can be used to let your objects appear and disappear as well as move on a finer grained scale.
+
+## Starting with the basics of SubAction
```
TheCedarPrince, line 19: Could you make each section title case?

```mkd
Starting with the basics of SubAction -> Starting with the Basics of `SubAction`
```

docs/src/tutorials/tutorial_4.md
TheCedarPrince: Perfect! Thanks Ole!

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,329 @@
...
+    [SubAction(20i+10:20i+20, Translation(0, 5)) for i in 1:5]...
+]),
+Action(60:150, (args...)->lip(lower_lip...); subactions=[
+    SubAction(1:15, appear(:fade)),
+    [SubAction(20i:20i+10, Translation(0, 5)) for i in 1:5]...,
+    [SubAction(20i+10:20i+20, Translation(0, -5)) for i in 1:5]...
+]),
+```
+
+Yeah I like those `...` splatting :smile:
```
TheCedarPrince, line 193: Can you replace with the correct emoji here?

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,329 @@
...
+
+To recap, by working through this animation you should now:
+
+1. Understand how to make objects appear and disappear using subactions
+2. Be able to move objects inside subactions to have a finer control of the movement
+3. Know how to use splatting (the `...`) to create actions in a loop
+
+Hope you had as much fun reading this tutorial as I had creating our mascot.
+
+You're now ready to create your own big project.
```
TheCedarPrince, line 233: Just fun, maybe put a clown emoji here at the end? :clown_face:

docs/src/tutorials/tutorial_4.md
```diff
@@ -0,0 +1,329 @@
...
+- a circular head
+- some hair
+- eyes
+- a nose
+- a moving mouth
+- he should say something
+
+## Learning Outcomes
+
+This tutorial should show you the power of subactions which can be used to let your objects appear and disappear as well as move on a finer grained scale.
```
TheCedarPrince, line 17: I would say that a better way to lay out the learning outcomes is like this:

```mkd
This tutorial demonstrates the power of subactions.
A `SubAction` can be used to finely manipulate objects in your animation or visualization. 
From this tutorial, you will learn how to:

1. To finely control objects by making them appear and disappear using a `SubAction`.

2. Move objects using a `SubAction`.

3. Learn Julia splatting syntax (the `...`) to create multiple actions quickly.
```

docs/src/tutorials/tutorial_4.md
Wikunia: Thanks I've changed parts of it as some were not correct but added your second paragraph as a summary.

The **note** should be `Action` not `SubAction`. We currently don't provide leaving the frames out for subactions.

docs/src/tutorials/tutorial_4.md
Wikunia: I have some font issues with some of those and wasn't able to find a one which works :smile:

docs/src/tutorials/tutorial_4.md
Wikunia: Good catch

Wikunia:
Hope this is good to go @TheCedarPrince Feel free to merge if ready!

TheCedarPrince:
Way to go @Wikunia :smile: !

PR #86, opened 2020-08-13
Tutorial on Actions

TheCedarPrince:
Ready for a first review. Let me know what you think!


docs/src/tutorials/tutorial_2.md
```diff
@@ -1,5 +1,367 @@
...
+                    "black",
+                    :fill,
+                    radius,
+                    electrode_names,
+                ),
+        ),
+...
+```
+
+Once all these modifications were made, execute your eeg code and you should get something that looks like this:
```
Wikunia, line 313: Would use `eeg` here

docs/src/tutorials/tutorial_2.md
```diff
@@ -1,5 +1,367 @@
...
+It's invocation in the code looks like this:
+
+```julia
+...
+Action(:same, :info, info_box),
+...
+```
+
+> **NOTE:** The function for `info_box` is a little different!
+> Each `Action` inside of the `javis` function automatically receives three objects being a `Video` object, which was previously defined outside of the `javis` function, the function, and the current frame number.
```
Wikunia, line 343: the `Action` object not the function :)

docs/src/tutorials/tutorial_2.md
```diff
@@ -1,5 +1,367 @@
...
+![](assets/eeg.gif)
+
+# Conclusion
+
+Congratulations! 🎉 🎉 🎉
+You made a brain! 
+To recap, by working through this animation you should now:
+
+1. Clearly understand how to use an `Action` and how to make your own `Action`
+2. See how to create complex animations
```
Wikunia, line 356: Be able to create your own or just see?

docs/src/tutorials/tutorial_2.md
```diff
@@ -1,5 +1,367 @@
...
+3. Know how to display meaningful information on your animations
+
+Great job leveling up your `Javis` skills! 💪
+
+---
+---
+
+> **Author(s):** Jacob Zelko, Ole Kröger \
+> **Date:** August 11th, 2020 \
```
Wikunia, line 365: 14th?

Shall we add this in every tutorial? Do we need that?

docs/src/tutorials/tutorial_2.md
```diff
@@ -1,5 +1,367 @@
...
+In this tutorial, we are going to learn how to make a brain! 🧠 
+Well, not _exactly_ making a brain. 
+Instead, we are going to animate brain activity by simulating a [10-20 EEG Electrode Array](https://en.wikipedia.org/wiki/10%E2%80%9320_system_(EEG)?oldformat=true) using random data. 
+This Project Tutorial builds on prior tutorials and serves as a cumulative test of your understanding about `Javis`.
 
+When you are done with this tutorial, you will have created the following animation:
 
+![](assets/eeg.gif)
+
+# Learning Outcomes
```
Wikunia, line 12: Maybe use `##`  here and for the following subsections. It should be smaller than the `# **Tutorial 2:** What Are Actions?` heading

docs/src/tutorials/eeg_animation.jl
```diff
@@ -0,0 +1,130 @@
+using Javis
```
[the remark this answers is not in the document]
TheCedarPrince: That needs to go into the JuliaTemplates repo... I am just going to make that quickly and add this in.

docs/src/tutorials/tutorial_2.md
TheCedarPrince: I would argue yes because the elastic search engine would pull up the tutorial based on these tags. What do you think?

docs/src/tutorials/eeg_animation.jl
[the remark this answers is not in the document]
TheCedarPrince: Done

docs/src/tutorials/tutorial_2.md
Wikunia: I think the wording is weird :stuck_out_tongue:

docs/src/tutorials/tutorial_2.md
```diff
@@ -1,5 +1,367 @@
...
+3. Know how to display meaningful information on your animations
+
+Great job leveling up your `Javis` skills! 💪
+
+---
+---
+
+> **Author(s):** Jacob Zelko, Ole Kröger \
+> **Date:** August 11th, 2020 \
+> **Tag(s):** brain, EEG, project, tutorial, electrodes
```
Wikunia, line 366: I would use tags about what the user learned like `actions` does that make sense?

docs/src/tutorials/tutorial_2.md
TheCedarPrince: Agreed. Fixed!

docs/src/tutorials/tutorial_2.md
TheCedarPrince: That makes complete sense to me. I added that in. :smile:

PR #107, opened 2020-08-20
Support all frame types in SubActions

Wikunia:
In general this way it is much cleaner and it does not matter in which order Action objects are created i.e
```
a = Action(...)
b = Action(...)
javis(v, [b, a])
```
works which didn't work before.


src/util.jl
```diff
@@ -0,0 +1,17 @@
+"""
+    compute_frames!(actions::Vector{AA}) where AA <: AbstractAction
+
+Set action.frames.frames to the computed frames for each action in actions.
+"""
+function compute_frames!(actions::Vector{AA}) where AA <: AbstractAction
+    last_frames = nothing
+    for action in actions
+        if last_frames === nothing && get_frames(action) === nothing
+            throw(ArgumentError("Frames need to be defined explicitly, at least for the first frame."))
```
TheCedarPrince, line 10: Could you make this ArgumentError a little more precise? I think what you could say is something like "Frames must be set explicitly in either the initial `Action` or `BackgroundAction` object." Wording is up to you but I find "at least for the first time" a little too vague.

Wikunia: How about: "Frames need to be defined explicitly in the inital AbstractAction like Action/BackgroundAction or SubAction."?

Wikunia:
To answer your question:
`Frames` can be defined by the user in three different ways:
- UnitRange
- Symbol
- Rel

The actual frames are `UnitRange` and they get computed inside the `javis` function.
Before this PR we used different constructors to achieve this but this is cumbersome if we have a lot of different AbstractActions.
This PR makes that very simple such that there is no special code change for SubAction besides that we have to call `compute_frames!` for each of them.  

Does that make more sense?

src/util.jl
TheCedarPrince: Sounds good to me!

TheCedarPrince:
@Wikunia - that makes a lot more sense to me now. Hmmm... This discussion might be worth adding into the tutorial we have on Actions. Or maybe to an advanced `Action` tutorial in the future.

Wikunia:
It's kinda outside the scope of tutorials, isn't it? I think it's only relevant whenever people want to join development

Wikunia:
So approved? :smile:

PR #108, opened 2020-08-20
use font size for latex

### LINKED ISSUE #23: fontsize and color in Javis instead of Luxor
Currently the `latex` function needs a font size as well as I have no clue whether I can get the font size from Cairo/Luxor. Maybe @cormullion knows :smile: 

Same is treu for color where it might be helpful to for #15 to morph one shape with a color into another shape with a different color. For this we would need to get the color set in the two function to blend between the two colors.

Wikunia:
It's now possible to use `fontsize` for latex and one can specify the position without the need to write `translate` before `latex`.
Fix #23 without using the new versions of Luxor or Cairo


src/Javis.jl
```diff
@@ -170,15 +170,17 @@ The current settings of an [`Action`](@ref) which are saved in `action.current_s
 - `opacity::Float64`: the current opacity
 - `mul_opacity::Float64`: the current multiplier for opacity.
     The actual opacity is then: `mul_opacity * opacity`
+- `fontsize::Float64` a
```
TheCedarPrince, line 173: I think you forgot the rest of the docstring here.

Wikunia: Oh yeah I had Real and thought Float64 is fine and then I forgot it. Good catch!

PR #112, opened 2020-08-20
Adding basic functionality for generating mp4 file

TheCedarPrince:
This was surprisingly easy to do. I found the solution [here](https://stackoverflow.com/questions/24961127/how-to-create-a-video-from-images-with-ffmpeg).

Anything else I should test on this?


src/Javis.jl
```diff
@@ -941,6 +941,12 @@ function javis(
         ffmpeg_exe(`-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png -i
                     "$(tempdirectory)/palette.bmp" -lavfi
                     "paletteuse=dither=sierra2_4a" -y $pathname`)
+    elseif ext == ".mp4"
+        # generate a colorpalette first so ffmpeg does not have to guess it
+        ffmpeg_exe(`-loglevel panic -i $(tempdirectory)/%10d.png -vf
```
Wikunia, line 946: is this necessary for mp4? At least in your next line this is not used.

src/Javis.jl
```diff
@@ -941,6 +941,12 @@ function javis(
...
+                    "palettegen=stats_mode=diff" -y "$(tempdirectory)/palette.bmp"`)
+	ffmpeg_exe(`-loglevel panic -i $(tempdirectory)/%10d.png -framerate $framerate -c:v libx264 -r 30 -pix_fmt yuv420p $pathname`)
+
     else
         @error "Currently, only gif creation is supported and not a $ext."
```
Wikunia, line 951: Please change this line to show that we support `.mp4` now :wink:

src/Javis.jl
```diff
@@ -941,6 +941,12 @@ function javis(
         ffmpeg_exe(`-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png -i
                     "$(tempdirectory)/palette.bmp" -lavfi
                     "paletteuse=dither=sierra2_4a" -y $pathname`)
+    elseif ext == ".mp4"
+        # generate a colorpalette first so ffmpeg does not have to guess it
+        ffmpeg_exe(`-loglevel panic -i $(tempdirectory)/%10d.png -vf
+                    "palettegen=stats_mode=diff" -y "$(tempdirectory)/palette.bmp"`)
+	ffmpeg_exe(`-loglevel panic -i $(tempdirectory)/%10d.png -framerate $framerate -c:v libx264 -r 30 -pix_fmt yuv420p $pathname`)
```
Wikunia, line 948: this line has a different indentation. Seems like that this occurs often. My vs code uses 4 spaces.

More importantly this does not seem to work for me. I had to change it to:
```
-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png -c:v libx264 -pix_fmt yuv420p $pathname
```
so `-framerate` needs to be before `-i` and I dropped `-r` as I think this might be confusing.

Wikunia (changes requested):
Some comments but in general it works. Biggest error is the framerate.

Additionally this feature should be tested. Maybe using: https://juliaio.github.io/VideoIO.jl/stable/reading/#VideoIO.get_duration

Project.toml
```diff
@@ -23,6 +23,7 @@ ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
 Images = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
 ReferenceTests = "324d217c-45ce-50fc-942e-d289b448e8cf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+VideoIO = "d6d074c3-1acf-5d4c-9a43-ef38773959a2"
```
Wikunia, line 27: I think there is a way to set a compat bound for test dependicies. As this is failing for an older version we should add that  😊

src/Javis.jl
```diff
@@ -941,8 +941,11 @@ function javis(
         ffmpeg_exe(`-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png -i
                     "$(tempdirectory)/palette.bmp" -lavfi
                     "paletteuse=dither=sierra2_4a" -y $pathname`)
+    elseif ext == ".mp4"
+	ffmpeg_exe(`-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png  -c:v libx264 -pix_fmt yuv420p $pathname`)
```
Wikunia, line 945: this is still wrong indented on GitHub and in VSCode

Wikunia: oh maybe you're aware of this :smile: . Thought you messaged me that it should be fine now.

TheCedarPrince: I think this is finally finally fixed! :joy: I checked it on Vim, VSCode, and plaintext! I firmly believe the indentation should be correct now!

Project.toml
TheCedarPrince: This is a great problem. I think I found a few different solutions thanks to @mkitti from Slack. Going to keep working on this for a bit. Will update in my next push for you to review.

TheCedarPrince:
Hey @Wikunia , I figured out how to ensure compat for testing based off this documentation here: https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1

The changes to the Project.toml for javis and the addition of the `test/Project.toml` account for this. Furthermore, testing passes on my machine and I explained the testing scheme for mp4's. Let me know what you think!

Project.toml
```diff
@@ -16,13 +16,3 @@ LaTeXStrings = "1.1"
 LightXML = "0.9"
 Luxor = "2"
 julia = "1.4"
-
-[extras]
```
Wikunia, line 20: Oh we don't need this if we have a `test/Project.toml`? Good to know.

test/animations.jl
```diff
@@ -35,83 +35,196 @@ end
...
+            ),
+            Action(
+                1:2,
+                (video, args...) -> path!(path_of_red, get_position(:red_ball), "red"),
+            ),
+            Action(:same, (video, args...) -> path!(path_of_blue, pos(:blue_ball), "blue")),
+            Action(1:2, (args...) -> rad(pos(:red_ball), pos(:blue_ball), "black")),
+        ],
+        tempdirectory = "images",
+        pathname = "dancing.mp4",
```
Wikunia, line 131: I thought we wanted to use a different frame rate here for testing?

Wikunia (commented):
Please have a look at the final comments and feel free to merge then.

Wikunia (approved):

Project.toml
TheCedarPrince: Yea! Isn't that cool??? Everything still works fine and it keeps it nicely organized. :)

Issue #92, opened 2020-08-15
Allow user to write loop over frames himself by providing a single frame render Javis like function


freemin7:
Currently Javis processes a set of actions over a certain set of frames. However for certain applications it might be preferable to give the user control over rendering loop if he wants to write into several files in parallel and it might help with consistent frame rates for live streaming too.

A Javis would be called and return after preparing/rendering a single/(fixed number of) frame. It would preferable if it also kept all the framebuffers etc allocated not unlike Integrator Interface from the DiffEq eco system. ( https://diffeq.sciml.ai/release-5.0/basics/integrator.html )

Wikunia:
Can you give an example on how one would use this functionality and why?
Is the problem mostly that Javis is kinda slow atm? Then #102 might solve this.

freemin7:
This could also be accomplished by providing the user a function (or parameter into the javis call) to render a certain frame of an animation.
This might be easier to implement, although it would occur all constant overhead that Javis has once per frame. I don't guess that will be a problem though.

TheCedarPrince:
Hey @freemin7 , I think you will have this functionality in PR #138 in the function `get_javis_frame` in v0.2.0. Let us know if that works for you!

TheCedarPrince:
We need to add a preprocess functionality such that a particular frame can be retrieved without having to preprocess all frames.

Wikunia:
Can this be closed as we have a preprocess function now?

TheCedarPrince:
Ping @freemin7 that this functionality has been added in v0.3.0 of Javis.

Closed! :smile:

freemin7:
Is there already some doc for it?

Wikunia:
It's not exported @freemin7 but you can have a look at the docstring with `? Javis.preprocess_frames!`

It basically computes the actual frames for each `Object` and `Action`.

Issue #91, opened 2020-08-14
Supporting live streaming like Manim


freemin7:
Manim has a `python -m manim --livestream` option and even options for Twitch `python -m manim --livestream --to-twitch --with-key [streamkey]` which allow to write the live rendered content to Twitch in real time.

It would be cool to be able to the same from Javis. This functionality might be better suited for a package on top of Javis.
As far as i understand ffmpeg provides that functionality.

freemin7:
The related issue in Manim and how it was added is documented here: https://github.com/3b1b/manim/issues/335

Wikunia:
Thanks for opening the issue. What I can see in the nearer future is the following:
We can directly display the current frame in a window instead of saving it into a file. Then we can use OBS to stream our screen. I know it's dull but at least we would not need to wait to create a `gif` or later `mp4` but instead could see directly what gets animated. 

In general I plan on using an images array instead of saving every frame to an image file as this is what currently slows down Javis.

TheCedarPrince:
Hey @freemin7, just following up here, any updates? I saw a post you made in the Julia Slack asking about `ffmpeg` regarding streaming:

```
ffmpeg -re -i input.mkv -c:v libx264 -preset veryfast -b:v 3000k -maxrate 3000k \
-bufsize 6000k -pix_fmt yuv420p -g 50 -c:a aac -b:a 160k -ac 2 \
-ar 44100 -f flv rtmp://live.twitch.tv/app/<stream key>
```

freemin7:
It might be interesting to have minimal examples like this:
 - A clock widget which streams the current system time (not a calculated time) and displays an image of an cuckoo between X:00:00 - X:00:59
 - If possible an interactive demo where you can set functions in a Julia shell to change what is displayed in real time

Both demos should work with OBS.

TheCedarPrince:
@freemin7 - the latter should be in our documentation. The former idea is being worked on in #345

I know you requested this feature a while ago but I would love to see what animations you should come up with! :)

freemin7:
I didn't find them a new one in https://wikunia.github.io/Javis.jl/dev/examples/#Examples .

https://wikunia.github.io/Javis.jl/dev/workflows/#Live-Streaming shows how to set it up with OBS. However i am unsure how to set up an background as 1:Inf seems dubious.

Wikunia:
Hi @freemin7 , 
it's unfortunately not yet live streaming see #345 
[...]

PR #119, opened 2020-08-25
[WIP] Live Preview

TheCedarPrince:
Initial ideas on creating a previewing system of animations and gifs via the usage of `Images.jl` and `ImageView.jl`.

Currently, there is no functional code ready for testing yet. I wanted to write down my thoughts before I forgot about them here:

My initial thought of handling this problem is to store the image in the buffer of one's computer. From there, preview the image by serializing it to some graphical viewer. Once the graphical viewer exits, the preview process finishes. Right now, for my proof of concept, I am constraining the problem as follows (lemme know your thoughts @Wikunia):

1. Create a method in the `javis` function to create images that are saved into the buffer when the method is called with something like `liveview::Bool`.

2. Determine how to convert the Luxor based `Drawing` object to an object compatible to a form suitable for being read by other viewing libraries.

3. Preview one image with one frame in a viewer. 

4. Close the viewer and have the script gracefully finish execution.

Some notes I have on the process so far.

- I inspected `Images.jl` and using the `load` function, the image is loaded in as the type: `Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}`. The `Luxor.jl` based `Drawing` object can be stored into the buffer using the call type: `Drawing(500, 500, :image)`

- There is a native function in `Luxor.jl` to get the matrix representation of the stored `Drawing` object called `image_as_matrix()` which must be called after defining everything to be drawn onto a `Drawing`. By the following `collect(image_as_matrix())` call, the returned matrix is of the form `Array{ARGB32,2} with eltype ARGB32`.

- An interesting dilemma is to determine whether or not @cormullion would want a native function in `Luxor.jl` to convert the `Drawing` object's matrix to a matrix type compatible with `Images.jl`. Or if we should make our own method for handling `Images.jl` type...

- I wonder what would be the best viewer for CLI based viewing. Looking at `GTK.jl` and `ImageView.jl` right now

- Fun fact, this should be able to be rendered natively in a Jupyter notebook without any extra work! :D 

- I wonder if we could make some sort of REPL based communication tool to the viewer to go forward with images, skip to a frame, etc. 

Further, I had a great conversation with @staticfloat on the Julia Slack about this interesting problem space. Here are excerpts before I lose them to the Slack void:

> That I don’t know; usually I do stuff like RGB.(x) to convert an array x from one type to an RGB type, but that depends on RGB having a constructor that looks like RGB(::ARGB32).

> If that doesn’t exist, you may be able to define something like ThrowAwayAlpha(x::ARGB32) = RGB(x.r, x.g, x.b) or something like that.

> then you can do ThrowAwayAlpha.(array)

**Additional Info about v0.2**
This PR introduces some extra dependencies therefore merging is done into `v0.2` and not `master`.


cormullion:
By spooky coincidence I've just been rewriting the imagematrix functions, because of bugs in transparency handling... My current code, on master and intended for a breaking Luxor version 3.0, outputs a plain array of arrays of floats (potentially RGBA values):

```
julia>  m = @imagematrix begin
               background(1, 1, 0, 0.5)
       end 2 2
2×2 Array{Array{Float64,1},2}:
 [1.0, 1.0, 0.0, 0.501961]  [1.0, 1.0, 0.0, 0.501961]
 [1.0, 1.0, 0.0, 0.501961]  [1.0, 1.0, 0.0, 0.501961]
```
 
The differences here are because arithmetic on the ARGB32 value `0xFFFF0080` is low precision. 

I'm just guessing at all this stuff - I'd rather be doing other things with Luxor. :)

Wikunia:
Can you @TheCedarPrince clarify what fails with `ARGB` seems like Images can handle that based on https://juliagraphics.github.io/Luxor.jl/stable/basics/#Drawing-as-image-matrix or am I wrong?

My initial thoughts on this were actually to not see one frame but all frames just live. 
In steps
1. We create an ImageView (or GTK) canvas
2. have a sleep statement depending on the frame rate
3. update the canvas for the next frame

The user might want to be able to specify something like rendered frames such that on a 10 min movie production it's easy to see the last 5 seconds as an example.

I also like the idea of jumping using the command line but I would move that to another issue for now.

Side note: I've seen that you changed formatting of the files which makes it harder to review your code in the future as it's unclearer which lines are actually effected. Do you mind to open a different PR for this formatting as I do think it looks nice :wink:

TheCedarPrince:
Can do @Wikunia ! My linter was a bit overzealous and decided to reformat everything. I will open a separate PR about syntax formatting since you like the style. I was using `JuliaFormatter.jl` for that formatting. Should work in VSCode as well!

Oh, nothing was failing, I was just writing up my notes/thoughts before I forgot them. I agree -- the CLI based jump tool is a separate thing altogether.

I agree with your vision as well for what could be done - I was just working on a proof of concept to get at least one frame shown. Your idea is the long-term goal for this PR. I typically solve complicated problems like this by breaking it up into smaller parts, solving each of those, and then recombining them to solve the big problem. I'll ping or Slack you as I make progress.

@cormullion spooky :ghost: ! I quite like that raw version of you handling the matrix. Seems like it would give greater flexibility to users so that they could map the returned type to whatever they need. Cool! Looking forward to v3.0! :smile:

TheCedarPrince:
I am not sure if `ImageView.jl` will work for our purposes. I just started a discussion in #image-processing in the Julia Slack about this strange behavior I am getting when I am attempting to preview a gif via `imshow`:

error

I will keep exploring to see what is possible!

Wikunia:
I would assume this is because of color palette issues. 
In our case I thought we would always just push the new image matrix and not a gif because in live view we would not create a gif.

TheCedarPrince:
That makes better sense @Wikunia . I guess what I am having trouble trying to figure out regarding an animation that has `1...n` number of frames and how to show specific frames to a user.

What I was thinking was that using the `image_to_matrix()` function, was to create a matrix of matrices where one could easily access frames they would want to see. The array I imagine could be of the form `n x length x width` where "length" and "width" are dimensions of the frame. If I want to view a subset of the frames, I could then access this matrix to get the specific frames I want to display.

However, where I get lost is how best to show not just one image, but any number of images. That was why I was exploring the `imshow` method as I thought that could potentially work for this case. It may still so I will continue examining it.

Do you understand where I am running into an issue? Feel free to Slack me if I am not being clear. If you have additional thoughts, let me know! Thanks!

Wikunia:
How about this? https://juliagizmos.github.io/GtkReactive.jl/stable/drawing.html
Being able to have a slider for the frame and the frame gets rendered and displayed when the user changes the slider.
Maybe we can also use: https://juliagizmos.github.io/GtkReactive.jl/stable/reference.html#GtkReactive.player not sure how to handle frame rate there but it might be an option.

cormullion:
@Wikunia don't know if [this](https://github.com/nodrygo/GtkLuxorNaiveDemo) is still useful

TheCedarPrince:
Thanks for the tip @cormullion ! I think this is the golden ticket for this problem!

I made some additional progress. I have this gif showing some of what I got working right now:

viewer_slider

What I was able to figure out was to get a slider to show that sends a signal to and from a textbox object. The textbox object can also send a signal back to the slider. It's pretty fun and what I imagine it being used for is via the slider or textbox, each time a new frame number is chosen, a callback action could take place to get the correct frame and render it in the viewer. My next step is to get an image to render in a GTK canvas, but I am struggling to know how to do it. Sadly, the documentation is not so great.

UPDATE: I got images working! Now to make the image appear with a slider/textbox widget...

TheCedarPrince:
This is ready for review @Wikunia. There are some notes and comments I left inside the `javis_viewer.jl` file so as to keep things in mind for future upgrades to the viewer. Let me know what you think! :smile:

Project.toml
```diff
@@ -15,7 +19,7 @@ VideoIO = "d6d074c3-1acf-5d4c-9a43-ef38773959a2"
 
 [compat]
 FFMPEG = "0.3, 0.4"
```
Wikunia, line 21: make sure to set upper bounds to the new `[deps]`

src/Javis.jl
```diff
@@ -1015,6 +1020,7 @@ function javis(
     actions::Vector{AA};
     framerate = 30,
     pathname = "javis_$(randstring(7)).gif",
+    liveview = false,
```
Wikunia, line 1023: this doesn't seem to be documented in the docstring

src/Javis.jl
```diff
@@ -1084,6 +1090,10 @@ function javis(
         end
         filecounter += 1
     end
+    
+    if liveview == true
+        javis_viewer(video = video, frames = length(frames), action_list = actions)
+    end
```
Wikunia, line 1096: I as a user expect that the live viewer shows up immediately and that no frames are getting rendered.
So `liveview=true` means: Showing the live viewer without rendering as we want to have a fast experience which was the main point for having an image viewer in the first place :wink:

src/javis_viewer.jl
```diff
@@ -0,0 +1,146 @@
+# TODO: Intelligent way of handling slider
+
+#=
+FIXME:
+Fascinating bug I found.
+Apparently to create a list of actions anywhere, you must first define a Video object.
+Else, you get an error like this:
+
+```
+ERROR: LoadError: BoundsError: attempt to access 0-element Array{Video,1} at index [1]
```
Wikunia, line 10: This is expected as an Action needs a Video. The error message should be improved though.

src/javis_viewer.jl
```diff
@@ -0,0 +1,146 @@
...
+
+    #####################################################################
+    # DISPLAY WIDGETS
+    #####################################################################
+
+    slide = slider(1:frames) # Creates a slider
+    tbox = GtkReactive.textbox(Int; signal = signal(slide)) # Creates a textbox
+
+    forward = GtkButton("==>") # Button for going forward through animation
+    backward = GtkButton("<==") # Button for going backward through animation
```
Wikunia, line 45: Not the best position for this comment but before I forget:
Can we make this to work on the left and right arrow button as I hate clicking :laughing:

src/javis_viewer.jl
```diff
@@ -0,0 +1,146 @@
...
+
+    #####################################################################
+    # VIEWER CANVAS AND GRID CONFIGURATION
+    #####################################################################
+
+    can = Gtk.Canvas(frame_dims[1], frame_dims[2]) # Frame size to be displayed.
+    grid = Gtk.Grid() # Grid to allocate widgets
+
+    # Allocate the widgets in the grid.
+    grid[1:3, 1] = can
```
Wikunia, line 63: Is this part of "To be or not to be?" like: "Can or can't we?" . Maybe call it canvas :wink:

Wikunia (changes requested):
Great progress so far!

As mentioned during the stream:
- It would be best to show the first frame by default

Some other things mentioned in the comments for specific lines.

src/javis_viewer.jl
TheCedarPrince: Opened a bug report on this guy #152 Feel free to close this convo.

src/javis_viewer.jl
TheCedarPrince: Got this fixed! :smile:

src/javis_viewer.jl
TheCedarPrince: Fair point! :joy:

src/Javis.jl
TheCedarPrince: Added to the docstring. Feel free to resolve the convo

Project.toml
TheCedarPrince: I reviewed the current package states and felt like I set reasonable boundaries with the versions. I decided upon the boundaries based on knowing when I started working with GTK and the liveviewer. So, at the current boundaries, everything works as expected. If you want me to revise these boundaries, let me know.

Wikunia: The boundaries and possibilities are way to specific. I actually wasn't able to `dev Javis`. It works only with Luxor 2.5.1 btw cormullion fixed the issue we had.
My new updated one is:
```
Animations = "0.4"
Cairo = "1.0.4, 1.0.5"
ColorTypes = "0.10"
FFMPEG = "0.3, 0.4"
Gtk = "1.1"
GtkReactive = "~1.0.3"
Images = "0.20, 0.21, 0.22"
LaTeXStrings = "1.1"
LightXML = "0.9"
Luxor = "2"
ProgressMeter = "1"
VideoIO = "0.6, 0.7, 0.8"
julia = "1.4"
```

Wikunia:
Great I resolved some conversations. Do you it is reasonable to make the viewer also loop from the last frame to the first one? This seems to make sense when one normally creates a `gif` at least :wink:

Wikunia:
Also format your code! Don't let your CI fail all the time :stuck_out_tongue:

Project.toml
TheCedarPrince: I adopted this scheme. Thanks for the help

src/Javis.jl
TheCedarPrince: This makes a lot of sense! I fixed this correctly now. :D

TheCedarPrince:
That makes complete sense @Wikunia ! I think this is finally ready for merge.

Wikunia:
I think we can't write tests for this as we can't open a display.

src/Javis.jl
```diff
@@ -1225,7 +1238,7 @@ end
 Get one specific frame of a video with actions.
 
 # Returns
-- `ARGB Matrix` the frame image as a matrix
+- `ARGB Matrix` - the frame image as a matrix
```
Wikunia, line 1241: Maybe use `typeof` to get the actual type or at least specify that it is two dimensional to be more exact?

src/javis_viewer.jl
```diff
@@ -0,0 +1,292 @@
...
+                    # Gets the correct Canvas context to draw on
+                    context = getgc(canvas)
+                    image(
+                        context,
+                        CairoImageSurface(frame_mat),
+                        0,
+                        0,
+                        frame_dims[1],
+                        frame_dims[2],
+                    )
```
Wikunia, line 202: this looks like it should go into a method as you use this more than once.

Wikunia: Probably put the whole `@guarded` one into a method and call with the `video, action_list, frame_number`

src/javis_viewer.jl
```diff
@@ -0,0 +1,292 @@
...
+                    0,
+                    frame_dims[1],
+                    frame_dims[2],
+                )
+            end
+        end
+    end
+
+    # When the `Right Arrow` key is pressed, increment current frame number
+    signal_connect(win, "key-press-event") do widget, event
```
Wikunia, line 168: this is the same as clicking the `==>` button, right? Therefore make it a method :wink:

src/javis_viewer.jl
```diff
@@ -0,0 +1,292 @@
+# TODO: Intelligent way of handling slider
+
+function javis_viewer(; video::Video, frames::Int, action_list::Vector)
```
Wikunia, line 3: The man who loves keyword arguments really uses them in weird places.
This will completely fail if you call `javis_viewer()` I suppose.

Wikunia (changes requested):
Great progress but yeah unfortunately there is still some stuff to do before merging.

Besides the comments on the specific lines you should make sure that the tests pass.

src/Javis.jl
```diff
@@ -1200,23 +1208,28 @@ function javis(
         filecounter += 1
     end
```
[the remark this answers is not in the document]
TheCedarPrince: Truth be told, I was rather uncertain about where `javis` does all the computation of the frames. This was a valuable lesson. I now fixed this problem and the live image viewer is quite fast!

src/Javis.jl
```diff
@@ -1200,23 +1208,28 @@ function javis(
...
-        ffmpeg_exe(`-loglevel panic -framerate $framerate -i $(tempdirectory)/%10d.png -i
-                    "$(tempdirectory)/palette.bmp" -lavfi
-                    "paletteuse=dither=sierra2_4a" -y $pathname`)
-    elseif ext == ".mp4"
-        finishencode!(video_encoder, video_io)
-        close(video_io)
-        mux("temp.stream", pathname, framerate; silent = true)
+    if liveview == true
+        javis_viewer(video = video, frames = length(frames), action_list = actions)
+        return "Live preview of $pathname"
```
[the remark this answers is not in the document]
TheCedarPrince: This has now been handled!

src/javis_viewer.jl
TheCedarPrince: Good point. I fixed this up. Let me know if you think I managed it properly.

src/javis_viewer.jl
TheCedarPrince: It has turned into a method!

src/javis_viewer.jl
TheCedarPrince: This is quite true... I turned this into a dunder function. I also like the nickname of "the man who loves keyword arguments".

src/Javis.jl
TheCedarPrince: This is updated and fixed.

TheCedarPrince:
All tests passed, I added thorough documentation to the functions and methods. Feel free to review @Wikunia.

src/Javis.jl
Wikunia: No it's still exactly the same :stuck_out_tongue:

src/javis_viewer.jl
Wikunia: not it hasn't. You still have twice the exact same code.

src/Javis.jl
```diff
@@ -1206,6 +1214,11 @@ function javis(
         CURRENT_ACTION[1] = actions[1]
     end
 
+    if liveview == true
+        _javis_viewer(video, length(frames), actions)
+        return "Live preview of $pathname"
```
Wikunia, line 1219: I think the `return` feels a bit wrong as `$pathname` doesn't exist anymore.

src/Javis.jl
```diff
@@ -1407,6 +1420,7 @@ export javis, latex
 export Video, Action, BackgroundAction, SubAction, Rel
 export Line, Translation, Rotation, Transformation, Scaling
 export val, pos, ang, get_value, get_position, get_angle
+export get_javis_frame
```
Wikunia, line 1423: I don't think we want to export this but if we do it needs better documentation like an example and explaining the arguments.

Wikunia (changes requested):
Sorry but it seems you missed quite a few of my review comments from last time. 
Again: If you use a code block of more than ~5 lines of code at least twice -> move it into a method

src/javis_viewer.jl
TheCedarPrince: Ah! I missed this. I have now properly modularized it

src/Javis.jl
TheCedarPrince: Fixed now.

src/Javis.jl
TheCedarPrince: Yea that makes sense. I just made the line simpler.

src/Javis.jl
TheCedarPrince: I actually do want to have this export as I think it would be useful to have. Plus, it has been somewhat requested on Zulip and in issues #91 and #92 . I don't know what a good example would be for this. Any ideas?

Wikunia: I think it actually can't be called when `javis` isn't called before. We should therefore have a preprocess  function which then can be exported as well. But not in this PR. For now I would not export it and the users can do Javis.get_javis_frame it show that this might change  in the future.

src/javis_viewer.jl
```diff
@@ -0,0 +1,230 @@
...
+    widgets::Vector,
+    actions::Vector,
+    dims::Vector,
+    canvas::Gtk.Canvas,
+    frames::Int,
+)
+    # Get current frame from textbox as an Int value
+    curr_frame = parse(Int, get_gtk_property(widgets[2], :text, String))
+    if frames > curr_frame
+        push!(widgets[1], curr_frame + 1)
```
Wikunia, line 43: Can you explain what widgets are so what is `widget[1]` and what is `widget[2]`?

src/javis_viewer.jl
```diff
@@ -0,0 +1,230 @@
...
+        slide_val = Gtk.get_gtk_property(bound_slide, "value", Int)
+
+        _draw_image(video, action_list, slide_val, canvas, frame_dims)
+    end
+
+    # When the `Enter` key is pressed, update the frame
+    signal_connect(win, "key-press-event") do widget, event
+        if event.keyval == 65293
+            # Get current frame from textbox as an Int value
+            curr_frame = parse(Int, get_gtk_property(tbox, :text, String))
```
Wikunia, line 186: I think `clamp` will solve this issue such that you don't need the `if elseif else` 
Btw I would call it `nframes` if it's the number of total frames

Wikunia: Or of course `total_nframes` :smile:

Wikunia (commented):
Great looks much cleaner! Maybe one or two things that can be improved.

src/javis_viewer.jl
TheCedarPrince: Added comments to explain what these two widgets represent

src/javis_viewer.jl
TheCedarPrince: I love `clamp`! Thanks for telling me about it! Also, changed it to total frames. :)

TheCedarPrince:
Hey! I reran the CI and things look OK @Wikunia ! Mind if I do the honors to merge it - if of course there are no more problems!

Wikunia:
Do you mind to also add a test for going from the last frame to the first frame and back?
Then it's good to go!

Wikunia (approved):
Looks good! Thanks so much! The rest goes into separate PRs

TheCedarPrince:
HELL YEA!!!!! :fire: :fire: :fire: Thanks for all the critiques! I am pretty happy with the first iteration of this viewer!

Issue #103, opened 2020-08-18
Transformation: Scale


Wikunia:
**Is your feature request related to a problem? Please explain.**
Javis support `Translation` and `Rotation` but not `Scaling` of objects

**Describe the solution you'd like**
Being able to animate scaling a circle as an example

PR #121, opened 2020-08-25
Scaling support

Wikunia:
## PR Checklist

- [X] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
   - Would be good to have a test case for scaling x and y differently
- [X] Did I check relevant tutorials that may be affected by changes in this PR?
   - [ ] A new tutorial might be reasonable or integrating it into the creation of Jarvis?

**Link to relevant issue(s)**

#103 

This PR contains several new functionality. It is now possible to write:
```
 video = Video(500, 500)
    javis(video, [
        BackgroundAction(1:50, ground),
        Action(:set_scale, (args...)->2),
        Action((args...)->circ(), subactions=[
            SubAction(1:15, Scaling(0.0, :set_scale)),
            SubAction(36:50, Scaling(:set_scale, 0.0)),
        ])
    ], tempdirectory="images", pathname=""
    )
```

This scales from 0.0 to `:set_scale` which equals 2 and then scales down again. 

It is equivalent to:

```
 video = Video(500, 500)
    javis(video, [
        BackgroundAction(1:50, ground),
        BackgroundAction(1:50, (args...)->scale(2)),
        Action((args...)->circ(), subactions=[
            SubAction(1:15, appear(:scale)),
            SubAction(36:50, disappear(:scale)),
        ])
    ], tempdirectory="images", pathname=""
    )
```

I needed to update `ActionSettings` to have a current scale the new desired scale and a multiplier. This is similar to how it's done with font size but the `current_scale` is needed as `Luxor.scale` is multiplicative such that `Luxor.scale(2); Luxor.scale(2);` is the same as `Luxor.scale(4)` which is obviously not the case for font size. Therefore, I keep track of the current scale and the one we want to have to compute the correct scale factor to make this change.

I also changed the `appear` and `disappear` methods as they contained a bug in the interpolation formula.

**Questions:**
Currently `Scaling(2.0)` is the same as `Scaling(1.0, 2.0)` but maybe it should be `Scaling(current_scale, 2.0)`. 
Scaling in general is a bit problematic as it's the global scale whereas sometimes one might want to have local scaling. What I mean is in Luxor the current scale is always 1.0 so when calling `scale(2)` everything gets bigger but calling it again makes everything twice as big again.

In Javis it's different:
```
BackgroundAction(1:50, ground),
Action(:set_scale, (args...)->2),
Action((args...)->circ(), subactions=[
    SubAction(1:15, Scaling(0.0, :set_scale)),
    SubAction(1:15, Scaling(0.0, :set_scale)),
    SubAction(36:50, Scaling(:set_scale, 0.0)),
])
```

is the same as 

```
BackgroundAction(1:50, ground),
Action(:set_scale, (args...)->2),
Action((args...)->circ(), subactions=[
    SubAction(1:15, Scaling(0.0, :set_scale)),
    SubAction(36:50, Scaling(:set_scale, 0.0)),
])
```


docs/src/tutorials/tutorial_4.md
```diff
@@ -62,7 +62,7 @@ Although one could use standard an `Action` to achieve the same functionality, `
 
 The `subactions` keyword uses a list of [`SubAction`](@ref) structs which are defined in a similar fashion as `Action` with `frames` and a `function` but are in some sense simpler than the `Action`. 
 
-A function of a `SubAction` is normally either [`appear`](@ref) or [`disappear`](@ref) at the moment or one of two transformations: [`Translation`](@ref) and [`Rotation`](@ref).
+A function of a `SubAction` is normally either [`appear`](@ref) or [`disappear`](@ref) at the moment or one of these transformations: [`Translation`](@ref), [`Rotation`](@ref) and [`Scaling`](@ref).
```
TheCedarPrince, line 65: Do you think we should include scaling in a separate tutorial since you never actually use scaling in the code of this tutorial?

src/Javis.jl
```diff
@@ -763,6 +817,26 @@ get_position(s::Symbol) = get_position(val(s))
 """
 pos(x) = get_position(x)
 
+# As it is just a number we can return it
+get_scale(x::Number) = x
+
+"""
+    get_scale(s::Symbol)
+
+Get access to the sclaing that got saved in `s` by a previous action.
```
TheCedarPrince, line 826: Typo: sclaing -> scaling

src/Javis.jl
```diff
@@ -763,6 +817,26 @@ get_position(s::Symbol) = get_position(val(s))
...
+- `Scaling`: the scale stored by a previous action.
+"""
+get_scale(s::Symbol) = get_scale(val(s))
+
+"""
+    scl(x)
+
+`scl` is just a short-hand for [`get_scale`](@ref)
+"""
+scl(x) = get_scale(x)
```
TheCedarPrince, line 838: Why do we want this short hand? I feel like the `scl` is not as explicit as the standard `get_scale`.

docs/src/tutorials/tutorial_4.md
Wikunia: Yes that would be nice. This PR is actually more complex than anticipated. I will add a tutorial in the next days once the functionality is good and approved.

src/Javis.jl
Wikunia: I like short hands :smile: I kind of enjoy that we have the 3 letter forms `pos`, `ang` and `scl`. Actually I'm not sure whether we need the `scl` and `get_scale` functionality as it basically saves just the val so we can access it with `val` / `get_value` and I don't have a good use case for it any way.

src/Javis.jl
```diff
@@ -219,16 +219,25 @@ The current settings of an [`Action`](@ref) which are saved in `action.current_s
...
 """
 mutable struct ActionSetting
     line_width::Float64
     mul_line_width::Float64 # the multiplier of line width is between 0 and 1
     opacity::Float64
     mul_opacity::Float64 # the multiplier of opacity is between 0 and 1
     fontsize::Float64
+    # scale has current scale, scale and scale multiplier
+    # such that scale*mul_scale is the desired scale
+    # the scale change needs to be computed using current_scale
```
TheCedarPrince, line 234: I am a little confused by these comments. What do you mean by "scale has current_scale"? I am overall just confused about what is stored where based on these comments. Could you perhaps clarify them or explain them to me in a different way?

src/Javis.jl
```diff
@@ -417,6 +429,10 @@ mutable struct InternalRotation <: InternalTransition
     center::Point
 end
 
+mutable struct InternalScaling <: InternalTransition
```
TheCedarPrince, line 432: You know, I have been thinking and it may be wise to actually explain each mutable struct that we define. The rationale for this is that if others were to hack on Javis, it may prove difficult to understand. Adding a small docstring for each one may be wise. Thoughts?

src/Javis.jl
TheCedarPrince: Haha, ok, ok, we can keep it. :wink: however, if you don't think it has any useful functionality, we can just remove it.

src/luxor_overrides.jl
```diff
@@ -46,13 +46,80 @@ function setopacity(opacity)
...
+Set the scale and multiply it with the current multiplier
+which is i.e. set by [`appear`](@ref) and [`disappear`](@ref).
+
+Normal behavior without any animation is the same as `Luxor.scale`.
+
+# Example
+```
+scale(0.5)
+circle(O, 20, :fill) # the radius would be 10 because of the scaling
+```
```
TheCedarPrince, line 101: Question: if one were to get the current value of the radius for the circle, would it return the 20 value or the 10 value? Where is the radius scaling getting mutated and stored?

src/luxor_overrides.jl
```diff
@@ -46,13 +46,80 @@ function setopacity(opacity)
...
+
+"""
+    scale(scl_x, scl_y)
+
+Same as [`scale`](@ref) but the x scale and y scale can be changed independently.
+
+# Arguments:
+- `scl_x`: scale in x direction
+- `scl_y`: scale in y direction
+"""
```
TheCedarPrince, line 118: Could you give an example of this? This is an interesting functionality that I am not sure if I understand how it works.

Wikunia: It just scales in two directions so it can stretch a square into a rectangle as an example.

src/luxor_overrides.jl
Wikunia: The whole canvas for this layer is scaled not the circle itself. So you can scale whatever you wish. The radius stays the same. Similar to doing a Translation.

src/Javis.jl
Wikunia: Scale saves the current scaling factor whereas the other two combined give the desired new scaling.

src/luxor_overrides.jl
TheCedarPrince: Ah gotcha. This makes sense.

src/luxor_overrides.jl
TheCedarPrince: Oh! That is nifty! Could you give an example of this? Thanks!

src/Javis.jl
Wikunia: We might want to have an actual development documentation for those as well but yeah as a start it might be good to have some docstrings.

src/luxor_overrides.jl
Wikunia: An example is inside the `test/animations.jl` named something like "squeezing" :smile:

src/Javis.jl
Wikunia: I think we do need them as scale supports a tuple and a single `Real` and needs to convert a single `Real` to a tuple.

docs/src/tutorials/tutorial_4.md
TheCedarPrince: Can you open an issue on this new tutorial and resolve this convo? Feel free to assign me the tut.

src/Javis.jl
TheCedarPrince: Sounds good. Resolved

src/Javis.jl
TheCedarPrince: That makes sense. Maybe this could be a separate issue? Feel free to open an issue and resolve this convo

Wikunia:
I'm not sure about `Grow` and `Shrink` but I think scaling from current scale is a reasonable default. Will add this and then I hope this is good to go. I've created #126 for the tutorial issue.

src/Javis.jl
Wikunia: Changed the comment

TheCedarPrince: Hey, I think somehow this comment got outdated. You still talk about scale although it is no longer used.

TheCedarPrince:
Ok, just one slight change on the documentation for ActionSetting. The comment you added seems to be outdated so it just needs a final update. Otherwise, I happily approve this PR and agree with you! @Wikunia

src/Javis.jl
Wikunia: Done

PR #125, opened 2020-08-27
Using Animations.jl

### LINKED ISSUE #42: Using Animations.jl
Regarding #22 where we discuss an easier way of defining frames.

Actually we might want to just define them as animations:
See https://jkrumbiegel.github.io/Animations.jl/stable/

What do you think about:
```
Action(anim, (args...)->line(O, Point(10,20), :stroke))
```

where anim is:
```
Animation(1, Translation(O), sineio(), 30, Translation(Point(20, 50))
```

but this might be a bit confusing so we should define short hands for it like

```
Action([1, Translation(O), sineio(), 30, Translation(Point(20, 50)], (args...)->line(O, Point(10,20), :stroke))
```

or 

```
Action(1:30, sineio, (args...)->line(O, Point(10,20), :stroke), Translation(O, Point(20, 50))
```

to define a simple animation.

Additionally I'm wondering whether the `Translation` should be in front of the function as this is the order of execution.

Wikunia:
## PR Checklist

   - Needs one for error checking
- [X] Did I properly add test dependencies to the `test` directory (if applicable)?
   - Added Animations.jl
   - Would be awesome if you can do this @TheCedarPrince 

**Link to relevant issue(s)**
Closes #42 

Currently the following syntax is used to define the easing function for interpolation:
- Use `linear` as default which makes this non-breaking
```
video = Video(1000, 700)
javis(
    video, [
        BackgroundAction(1:150, ground),
        Action((args...)->circle(Point(-300, -300), 25, :fill); subactions=[
            SubAction(1:50, sineio(), Translation(500, 0))
            SubAction(51:100, polyin(2), Translation(0, 500))
            SubAction(101:150, expin(8), Translation(-500, -500))
        ])
    ], pathname="current/sine_trans.gif"
)
```
with the result: sine_trans

or similarly:
```
javis(
    video, [
        BackgroundAction(1:150, ground),
        Action((args...)->circle(Point(-300, -300), 25, :fill), sineio(), Translation(500, 0))
    ], pathname="current/sine_trans.gif"
)
```
if one doesn't work with subactions.

The appear, disappear and transformation functions now access the interpolation value `t` via `get_interpolation(action, frame)`.

This does currently not contain tests but contains a bugfix which is also available in #121 which defines the linear interpolation in appear and disappear which was wrong before. 

I would like to be able to write something like:
```
circle_anim = Animation(
    [0, 0.3, 0.6, 1], # must go from 0 to 1
    [O, Point(500, 0), Point(500, 500), O],
    [sineio(), polyin(5), expin(8)],
)
javis(
    video, [
        BackgroundAction(1:150, ground),
        Action((args...)->circle(Point(-300, -300), 25, :fill); subactions=[
            SubAction(1:150, circle_anim, translate())
        ])
    ], pathname="current/sine_trans.gif"
)
```

this will be done in the next step. Anything against this syntax? @TheCedarPrince 

**Additional Info about v0.2**
This PR introduces `Animations.jl` therefore merging is done into `v0.2` and not `master`.


Wikunia:
The last commit makes the following reality:

A section from the docs:

    translate()

Translate a function defined inside an [`Action`](@ref) using an Animation defined
with Animations.jl.

If you're used to working with Animations.jl this should feel quite natural. 
Instead of defining each movement in its own subaction it's possible to define it in one
by using an Animation.

# Example
```
using Javis, Animations

function ground(args...)
    background("black")
    sethue("white")
end

video = Video(500, 500)
circle_anim = Animation(
    [0.0, 0.3, 0.6, 1.0], # must go from 0 to 1
    # the circle will move from the origin to `Point(150, 0)` then `Point(150, 150)`
    # and back to the origin `O`.
    [O, Point(150, 0), Point(150, 150), O],
    [sineio(), polyin(5), expin(8)],
)
javis(
    video, [
        BackgroundAction(1:150, ground),
        Action((args...)->circle(O, 25, :fill); subactions=[
            SubAction(1:150, circle_anim, translate())
        ])
    ], pathname="moving_a_circle.gif"
)
```

Here `circle_anim` defines the movement of the circle. The most important part is that the
time in animations has to be from `0.0` to `1.0`.

src/Javis.jl
```diff
@@ -140,13 +142,47 @@ A SubAction should not be created by hand but instead by using one of the constr
...
+    ])
+])
+
+# Arguments
+- `frames`: A list of frames for which the function should be called.
+    - The frame numbers are relative to the parent [`Action`](@ref).
+- `easing::Easing`: The easing function for `args...`
+- `args...`: Either a function like [`appear`](@ref) or a Transformation
+    like [`Translation`](@ref)
+"""
```
TheCedarPrince, line 177: Good job on the docstring! I would suggest changing the format and presentation like this:

```mkd
A `SubAction` can be defined with frames, an easing function, and either a function or
transformation(s).

# Arguments
- `frames`: A list of frames for which the function should be called.
    - The frame numbers are relative to the parent [`Action`](@ref).
- `easing::Easing`: The easing function for `args...`
- `args...`: Either a function like [`appear`](@ref) or a Transformation
    like [`Translation`](@ref)

# Example
In the following example a filled circle with radius 50 appears in the first 20 frames,
which means the opacity is increased from 0 to 1.0.
The interpolation function used here is
sineio from [`Animations.jl`](https://github.com/jkrumbiegel/Animations.jl).
Then it stays at full opacity and disappears in the last 20 frames using a linear decay.

```julia
javis(demo, [
    BackgroundAction(1:100, ground),
    Action((args...)->circle(O, 50, :fill); subactions = [
        SubAction(1:20, sineio(), appear(:fade)),
        SubAction(81:100, disappear(:fade))
    ])
])
```


Also, it might be good to also explain what an easing function is and link to the easing functions available at `Animations.jl`.

src/Javis.jl
```diff
@@ -358,6 +401,41 @@ Similar to the above but uses the same as frames as the action above.
 Action(func::Function, args...; kwargs...) =
     Action(:same, nothing, func, args...; kwargs...)
 
+"""
+    Action(frames, id::Union{Nothing,Symbol}, func::Function,
+           transitions::Transition...; kwargs...)
+
+Fallback constructor for an Action which doesn't define an animation.
+A linear animation is assumed.
+"""
```
TheCedarPrince, line 410: Perhaps also add an example here similar to what you did for the constructor that you made for an Action which does define an animation using an easing function.

src/Javis.jl
```diff
@@ -1,5 +1,6 @@
 module Javis
 
+using Animations
```
TheCedarPrince, line 3: I also think it might be wise to add in automatic exports of `Animations.jl` methods/objects similar to what we did for `Luxor.jl`. I don't like having to state at the beginning of Javis files `using Javis, Animations`. I just want to use Javis! :smile:

src/subaction_animations.jl
```diff
@@ -67,13 +65,61 @@ function disappear(s::Symbol)
...
+    [sineio(), polyin(5), expin(8)],
+)
+javis(
+    video, [
+        BackgroundAction(1:150, ground),
+        Action((args...)->circle(O, 25, :fill); subactions=[
+            SubAction(1:150, circle_anim, translate())
+        ])
+    ], pathname="moving_a_circle.gif"
+)
```
TheCedarPrince, line 111: One thing that I am somewhat nervous about is the fact that there is two different syntax types we offer: `Javis.jl` Action syntax and `Animations.jl` Animation syntax. I think the way you could fix this is to show both versions of the syntax. Here is the syntax for Javis Actions that achieves the same animation:

```julia
javis(                                                                                 
    video,                                                                             
    [                                                                                  
        BackgroundAction(1:150, ground),                                               
        Action((args...)->circle(O, 25, :fill); subactions = [                         
            SubAction(1:50, sineio(), Translation(150, 0)),                            
            SubAction(51:100, polyin(2), Translation(0, 150)),                         
            SubAction(101:150, expin(8), Translation(-150, -150))                      
        ])                                                                             
    ],                                                                                 
    pathname = "moving_a_circle_javis.gif",                                            
)                              
```

src/subaction_animations.jl
```diff
@@ -67,13 +65,61 @@ function disappear(s::Symbol)
...
+```
+
+Here `circle_anim` defines the movement of the circle. The most important part is that the
+time in animations has to be from `0.0` to `1.0`.
```
TheCedarPrince, line 115: I'm a little confused as to why it must start at 0. Is that to say, this is where the circle or shapes originate?

src/util.jl
```diff
@@ -26,3 +26,17 @@ function get_current_setting()
...
+    t = min(1.0, t)
+end
+
+function get_interpolation(action::AbstractAction, frame)
+    t = get_interpolation(get_frames(action), frame)
+    if !(action.anim.frames[end].t ≈ 1)
+        @warn "Animations should be defined from 0.0 to 1.0"
+    end
+    return at(action.anim, t)
+end
```
TheCedarPrince, line 42: Could you add a small docstring here similar to what you did for `get_current_setting()`? Thanks!

src/Javis.jl
Wikunia: Normally we have Arguments at the end, right? Do you think this order is better or shall I just update the explanation?

src/Javis.jl
Wikunia: I think the user needs to understand once how to define a general Action which I think should be part of our documentation. Does that makes sense? I don't think it's really useful to provide an example for each of them as users might ask: Why can `id` be `nothing` here and do I need to define it? It's basically just to limit the number of functions internally to express all possibilities but not actually what the user calls. It's often the case that one constructor calls one method wich calls another constructor.

src/subaction_animations.jl
Wikunia: Why does that make you nervous? A lot of stuff can be done in different ways :smile: But yeah I can add that to the docstring

src/Javis.jl
Wikunia: Short recap of my reasoning:
I think we should acknowledge with `using Animations` that this is part of their package. In general I think it's nice to see for the user where stuff is coming from. Luxor is an exception as we overload the functionality so reexporting is our only option.

src/Javis.jl
TheCedarPrince: Hmmm... I missed that we switched around arguments and examples. Now that I think about it, your organization scheme I think is better with Examples first and the Arguments later. Just update the explanation and otherwise, this is fine.

src/Javis.jl
TheCedarPrince: That makes perfect sense. Feel free to close this convo

src/Javis.jl
TheCedarPrince: Agreed!

src/subaction_animations.jl
TheCedarPrince: Thanks! Feel free to resolve this once you are done with adding the docstring. I guess my nervousness has to do with the fact that we do not control Animations.jl, therefore, the syntax could change at any moment and would leave us scrambling. We control Javis syntax so I have no concerns there.

src/Javis.jl
Wikunia: Hope the docstring is better now.

Wikunia:
If you think the `translate` function makes sense I would add the same for `scale` and `rotate`.

TheCedarPrince:
I think the `translate` function makes sense so feel free to add the other two @Wikunia! Otherwise, I just finished doing a review this and you have my approval to merge when ready.

Wikunia:
animations

Can be done with:

```julia
using Javis, Animations

function ground(args...)
    background("black")
    sethue("white")
end



video = Video(500, 500)
translate_anim = Animation(
    [0, 1], # must go from 0 to 1
    [O, Point(150, 0)],
    [sineio()],
)

translate_back_anim = Animation(
    [0, 1], # must go from 0 to 1
    [O, Point(-150, 0)],
    [sineio()],
)

rotate_anim = Animation(
    [0, 1], # must go from 0 to 1
    [0, 2π],
    [linear()],
)

javis(
    video,
    [
        BackgroundAction(1:150, ground),
        Action(
            (args...) -> circle(O, 25, :fill);
            subactions = [
                SubAction(1:10, sineio(), scale()),
                SubAction(11:50, translate_anim, translate()),
                SubAction(51:100, rotate_anim, rotate_around(Point(-150, 0))),
                SubAction(101:140, translate_back_anim, translate()),
                SubAction(141:150, rev(sineio()), scale()),
            ],
        ),
    ],
    pathname = "current/animations.gif",
)

```

src/subaction_animations.jl
```diff
@@ -73,19 +70,190 @@ function disappear(s::Symbol)
...
+    video, [
+        BackgroundAction(1:150, ground),
+        Action((args...)->circle(O, 25, :fill); subactions=[
+            SubAction(1:150, circle_anim, translate())
+        ])
+    ], pathname="moving_a_circle.gif"
+)
+```
+
+This notation uses the Animations.jl library very explicitly. It's also possibble to do the
```
TheCedarPrince, line 124: Typo: possibble -> possible

Wikunia:
Oh I think I should add some more coverage before merging and it would be nice to also have 
this functionality for `sethue` to interpolate colors. @experimentMartin wants to see more weird colors on my stream.

Wikunia:
animations

This is what we can create now with `sethue`.

TheCedarPrince (approved):
Everything works on my end as expected. Feel free to merge when ready!

PR #162, opened 2020-09-11
ability to animate text

### LINKED ISSUE #161: Draw a text
It's not really a problem but it would be nice to have something more visual appealing.

I would like to have some kind of interesting appearance of text from left to right. Based on #43 it is currently possible to let text of everything else for that matter fade in. Would be awesome to have `appear(:draw)` or something which only works for `text` at first and might make sense for everything else later.

Currently there is only the way to `fade` or `scale` but it doesn't show that text is from left to right (at least in English :smile: )

**Additional context**
Can be seen in manim here:
https://youtu.be/fNk_zzaMoSs?t=6 
Easier to see with 0.25x speed. It doesn't have to be this animation. A simple one like appearing from left to right using clipping might be a good start.

Wikunia:
## PR Checklist

  - Might be possible to integrate this into the Jarvis tutorial
  - or into the PeriodicTable one @TheCedarPrince Like showing the element in big above the circle?

**Link to relevant issue(s)**
Closes #161 

Text can now be animated using the `appear` function and the symbol `:draw_text`. 
Only needed to override the `text` function and added one `appear` and one `disappear` function.
```julia
using Javis, Animations

function ground(args...)
    background("black")
    sethue("white")
    fontsize(20)
end

video = Video(400, 300)

javis(
    video,
    [
        BackgroundAction(1:100, ground),
        BackgroundAction(1:100, (args...) -> fontsize(30)),
        Action(
            1:100,
            (args...) -> text("Hello Stream!", -50, 50; halign = :center);
            subactions = [
                SubAction(1:15, sineio(), appear(:draw_text)),
                SubAction(76:100, sineio(), disappear(:draw_text)),
            ],
        ),
    ],
    tempdirectory = "current/images",
)
```

text


Wikunia:
This also need some docstrings!

Wikunia:
This fails because text is displayed differently on different operating systems. We might wanna write some kind of unit test instead and check that it just doesn't throw an error.

src/luxor_overrides.jl
```diff
@@ -148,3 +148,89 @@ function scaleto(x, y)
...
 end
+
+function animate_text(
+    str,
+    pos::Point,
+    valign::Symbol,
+    halign::Symbol,
+    angle::Float64,
+    t::Float64,
+)
```
TheCedarPrince, line 159: Missing a docstring here -- I see a lot of the code came directly from Luxor. It seems like there is enough of a change to merit a docstring for the user.

TheCedarPrince (approved):
Passed all tests on my end. I think you are good to go here after fixing up a missing docsstring

src/luxor_overrides.jl
Wikunia: This is an internal function though. I'll add a short developer docstring.

TheCedarPrince (approved):
Thanks for fixing up the docstring some more. I think it looks much better and ready for merging. Great work @Wikunia!

PR #175, opened 2020-09-14
[WIP] Feature transpose

### LINKED ISSUE #16: Animate transposing a matrix
Came as a suggestion of @xiaodaigh during the stream today.

Wikunia:
## PR Checklist

  I think we don't need one for statistics and maybe I remove that anyway :smile: 

**Link to relevant issue(s)**
Closes #16 

This is a big PR that is completely work in progress. 
What I have done so far:
- morphing can be done with vector of vector of points now as well as with functions
- it only works if the start state has the same number of polygons as the end state
- currently tries to move objects that appear in the end state somewhere else instead of morphing
   - makes sense for transposing a matrix
- cleaned up some morphing stuff with using `polysample`, `mod1` and `circshift`

## This is work in progress so the example might fail from time to time :smile: 

**Example**
```julia
using Javis, Latexify, LaTeXStrings

function ground(args...)
    background("black")
    sethue("white")
    fontsize(50)
end

function matrix(; do_transpose = false, action = :stroke)
    # needs to be set here because I return the path. Maybe I should disallow this behavior again
    # maybe only (args...)->matrix() should be allowed
    fontsize(50)
    mat = ["α" "β" "γ"; "x^2" "sqrt(y)" "λ"; 1 2 "y"]
    if do_transpose
        mat = permutedims(mat)
    end
    str = latexstring(latexify(mat))
    action == :path && newpath()
    latex(str, Point(-200, -100), action)
    return pathtopoly()
end

function T()
    fontsize(40)
    latex(L"$T$", Point(180, -120), :stroke)
end

video = Video(600, 400)
javis(
    video,
    [
        BackgroundAction(1:120, ground),
        Action(1:30, (args...) -> matrix()),
        # currently the transpose T is hard-coded
        Action(
            1:120,
            (args...) -> T();
            subactions = [
                SubAction(31:60, disappear(:fade)),
                SubAction(91:120, appear(:fade)),
            ],
        ),
        Action(
            31:60,
            morph(
                matrix(; action = :path),
                matrix(; action = :path, do_transpose = true);
                action = :fill,
            ),
        ),
        Action(61:90, (args...) -> matrix(; do_transpose = true)),
        Action(
            91:120,
            morph(
                matrix(; action = :path, do_transpose = true),
                matrix(; action = :path);
                action = :fill,
            ),
        ),
    ],
    pathname = "current/transpose.gif",
)

```

produces
transpose

Note that the `T` is hard coded at the moment.

## This adds a new dependency so can be only inside of v0.3
Unfortunately it's not good enough for the upcoming v0.2 :confused: We might release v0.3 not too late afterwards if this passes all tests.


Wikunia:
Just needs some final clean up:
- more test cases
- reorder functions and add documentation
- remove unused functions

src/Shape.jl
```diff
@@ -0,0 +1,275 @@
...
+    perc = isnan(perc) ? 1.0 : perc
+    # println("Point score: ", perc*score_points)
+    score += perc * score_points
+
+    # number of holes
+    nA = length(shapeA.subpaths)
+    nB = length(shapeB.subpaths)
+    perc = nA < nB ? nA / nB : nB / nA
+    perc = isnan(perc) ? 1.0 : perc
+    # println("Hole score: ", perc*score_holes)
```
TheCedarPrince, line 119: I see that there a lot of `println` statements commented out in this code. Could you remove them if they are no longer needed?

src/Shape.jl
```diff
@@ -0,0 +1,275 @@
...
+    points::Vector{Point}
+    simplified_points::Vector{Point}
+    centroid::Point
+    centered_points::Vector{Point}
+    subpaths::Vector{Vector{Point}}
+    num_acute_angles::Int
+    num_obtuse_angles::Int
+    num_right_angles::Int
+end
+
```
TheCedarPrince, line 11: Everything here is great, but to be honest, I have no idea what the general purpose of these functions are. Could you add small dev docstrings to them so I could better review and understand what each one is doing? Thanks!

src/morphs.jl
```diff
@@ -102,39 +104,265 @@ circle_obj = Object(11:20, acirc)
...
+    bool_move = ones(Bool, number_of_shapes)
+    bool_appear = zeros(Bool, number_of_shapes)
+
+    for si in 1:number_of_shapes
+        from_shape = action.defs[:from_shape][si]
+        to_shape = action.defs[:to_shape][si]
+        inter_shape = action.defs[:inter_shape][si]
+
+        # polygons[pi] = from_poly
+        # println("#Points: $pi ", length(polygons[pi]))
```
TheCedarPrince, line 173: Are these lines needed?

src/morphs.jl
```diff
@@ -102,39 +104,265 @@ circle_obj = Object(11:20, acirc)
...
+    end
+
+    # let new paths appear
+    t = get_interpolation(action, frame)
+    setopacity(t)
+
+    shape_ids = [si for si in 1:number_of_shapes if (!bool_move[si] && bool_appear[si])]
+    shapes = action.defs[:inter_shape][shape_ids]
+    # print_basic.(shapes)
+    # error(1)
```
TheCedarPrince, line 199: Are these lines needed?

src/morphs.jl
```diff
@@ -102,39 +104,265 @@ circle_obj = Object(11:20, acirc)
...
+    for i in length(from_polys):-1:1
+        length(from_polys[i]) <= 1 && splice!(from_polys, i)
+    end
+    for i in length(to_polys):-1:1
+        length(to_polys[i]) <= 1 && splice!(to_polys, i)
+    end
+
+    # println("Number of non empty polygons")
+    # println("#From ", length(from_polys))
+    # println("#To ", length(to_polys))
```
TheCedarPrince, line 239: Are these lines needed?

TheCedarPrince (changes requested):
It's a little difficult to parse exactly what is happening with all the code. We may have to have a bigger chat about this so I can understand what is happening. Otherwise, from a high level, it makes sense with how you are trying to interpolate between each individual point. Also, all the tests pass and the final outcome is not so bad I think for this first version of morphing! Great stuff!

src/Shape.jl
Wikunia: sure

TheCedarPrince (approved):

PR #188, opened 2020-09-18
[Bugfix] return 1.0 in interpolation for single frame

### LINKED ISSUE #187: [BUG] Scaling only works in the origin. Can't move to origin by `Translation`
**Describe the bug**

Currently scaling does only work on objects that are centered. This makes sense as scale is performed on the canvas. A similar problem can be experienced with rotation.

**To Reproduce**

1.5.1

Linux

3. Minimum working code example that led to bug:

```julia
Action(1:100, (args...)->circle(Point(10, 20), 10, :fill); subactions= [
   SubAction(1:10, appear(:scale)
])
```
does not work as one would expect.

Expected: The circle scales from a 0 radius to the given radius of ten.
Would actually appears and makes total sense but is nonetheless unexpected:
The circle appears small in the center and grows and translates to the point.

One way to fix this which currently fails but should be addressed is:

```julia
Action(1:100, (args...)->circle(O, 10, :fill); subactions= [
   SubAction(1:1, Translation(Point(10, 20))),
   SubAction(1:10, appear(:scale)
])
```

this way the canvas gets shifted and scaling works just fine. This results in a problem in get interpolation caused by a divide by 0.
I suggest interpolation in a range of 1 should always return 1.0

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #187

If `get_interpolation` is called with the last frame it returns 1.0 without any computation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   - Currently must be invoked after `javis` function call
   - Can be called via `Javis.get_javis_frame` as it is not exported yet
 - An object described by an action can follow a path (a vector of points). See `follow_path`
+- Bugfix in interpolation: Interpolation of a single frame like `1:1` returns `1.0` now instead of `NaN`.
 
   
 ## 0.1.5 (14th of September 2020)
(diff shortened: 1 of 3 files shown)


TheCedarPrince (approved):
Approved - feel free to merge

PR #191, opened 2020-09-18
[Bugfix] scaling to 0

### LINKED ISSUE #190: [BUG] scaling to 0.0 in disappear breaks everything
**Describe the bug**

When having several actions and one of them disappears the scale value would hit `0.0` and that breaks Cairo.
See https://github.com/JuliaGraphics/Luxor.jl/issues/114

**To Reproduce**

1.5.1

Linux

3. Minimum working code example that led to bug:
```julia
using Luxor

@png begin
    @layer begin
        scale(0.0)
        circle(O, 50, :fill)
    end
    @layer begin
        circle(Point(100, 100), 50, :fill)
    end
end
````

yes that is Luxor code but it shows the result faster :smile: 

The expected behavior when scaling to 0.0 is that the object that gets scaled disappears but instead all objects disappear into deep black hole.

Bugfix is in the works by never scaling to 0.0 and instead set a variable `show_action=false` inside the current settings. Then the action function itself doesn't get called.

Wikunia:
## PR Checklist

  - can you double check that the periodic table tutorial still works? @TheCedarPrince 

**Link to relevant issue(s)**
Closes #190 

As Cairo doesn't really support scaling to `0.0` I set a flag in the current settings that the action should not be called when the scale is 0.0


TheCedarPrince (approved):
Thanks for clarifying the docstring on `show_action`. I appreciate the clarification about how if scaling like `scale(0, 0)` will cause everything to disappear but Cairo does not complain.

Issue #168, opened 2020-09-13
Update Tutorial 5 to Include Text Drawing


TheCedarPrince:
Upon completion of #161 , tutorial 5 needs to be updated to include appearing and disappearing of text.

PR #197, opened 2020-09-22
Animation showcase for Javis

### LINKED ISSUE #181: Create a new gallery showcase gif
For v0.2 we would like to showcase the new functionality on the Readme.

One way I can image is: instead of having a big gif as we currently have it bit be reasonable to have a table (without borders) to show the same thing but the gifs can be clickable to the tutorial where this explained or if there doesn't exist such a tutorial we can have an examples folder with a file for each animation on how to reproduce this.
This is way easier than writing a tutorial each time and should be sufficient after we explained the concepts of Javis.

What do you think @TheCedarPrince ?

TheCedarPrince:
**Link to relevant issue(s)**
Closes #181 

First draft of ideas for the showcase to be put into the README.


README.md
```diff
@@ -17,7 +17,21 @@ Check out our [documentation](https://wikunia.github.io/Javis.jl/dev/) for tutor
 
 ## Gallery
 
-![](assets/showcase.gif)
+### Tutorials
+
+| Animation                  | Animation                          |
```
Wikunia, line 22: Can we remove the unnecessary header?

Wikunia: Like maybe a good old fashion html table without borders :wink:

TheCedarPrince: I can't remove the header fully. I don't really want to mix HTML with markdown as I really don't like working with HTML if I can help it.

TheCedarPrince:
@Wikunia - could you merge #198 so I can add your car example to the "Examples" section?

README.md
Wikunia: Hmm that leave it as it is (revert the last commit as I think that looks odd)

Wikunia:
Is merged. As this is only affecting the readme it doesn't need to be perfect before releasing v0.2.
=> #168 is more important

Wikunia:
If you include #198 this can be merged as well.

PR #198, opened 2020-09-22
Wik examples follow path

Wikunia:
## PR Checklist

A first example for follow a path of a bezier curve without the need of a tutorial.

follow_bezier_path


TheCedarPrince (approved):
Great example! Could you go ahead and merge this one so I can add it to the examples in the showcase PR? Thanks!

PR #199, opened 2020-09-23
convert frame only once

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
None

Currently the `get_javis_frame` function would convert the frame to `ARGB32` and inside `javis` it would be converted to `RGB`. This is stupid :smile:  -> convert it directly to the image type we need.


TheCedarPrince (approved):
Approved -- no concerns on my end.

PR #203, opened 2020-09-25
Added Examples Page

TheCedarPrince:
I added an examples page. Right now the link for the follow-path example does not work yet as it has not been merged to master. When it has been though, the link should be fine.

@Wikunia - for some reason I am having difficulty displaying a gif inside the table. Could you take a look at what is going on and why I cannot see the gif being rendered? Thanks!


docs/src/examples.md
```diff
@@ -0,0 +1,10 @@
+## Examples
+
+Aside from tutorials which take you step by step on how to create a tutorial, we also felt that it useful to show additional examples on things you can make with `Javis`!
+These animations do not have an accompanying tutorial but rather, scripts that you can examine, tweak, and modify to suit your own purposes.
+If you have an interesting example that you would like to share with us, open an issue (check out our [Contributing Guide](contributing.md) on how to do this) and we'd be happy to add your example to this page!
+
+| Example                                                                                            |
+|-------------------------------------------------------------------------------------------------------------------------------------------------------|
+| ![](assets/follow\_bezier\_path.gif                                                                |
```
Wikunia, line 9: It seems like your missing a `)` here.

TheCedarPrince: :man_facepalming:

Wikunia:
As mentioned I would make this with three columns and one line per example.
The columns are gif, link and used functions. In this case I would mention: `follow_path, makebezierpath, appear(:scale)`
Don't you think that this will be helpful if people search for `follow_path` in the docs?
Maybe because it's not much text it can be:
Two column layout: Gif in one and short explanation of what is done with the link and those tag words in the other.

Explanation would be:
Creating a bezier curve to draw a car and let a circle follow a long that path. The animation uses also appear and disappear functionalities to make the animation more pleasing to watch. Blablabla :smile:

docs/src/examples.md
```diff
@@ -4,7 +4,6 @@ Aside from tutorials which take you step by step on how to create a tutorial, we
...
+|------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| ![](assets/follow_bezier_path.gif) | [Follow a Path](https://github.com/Wikunia/Javis.jl/blob/master/examples/follow_path.jl) | This creates a bezier curve to draw the outline of a car and have a circle go around that outlined path. [`follow_path`](@ref), [`makebezierpath`](@ref), and [`appear`](@ref) were used to create this animation. |
```
Wikunia, line 9: Makebezierpath is not available for ref as this is part of Luxor

Wikunia, line 9: Makebezierpath is not available for ref as this is part of Luxor

Issue #179, opened 2020-09-17
Discussion: Syntax for SubAction


xiaodaigh:
I find the current syntax for subaction non-composable. I think of subactions as just actions on an object.

**Describe the solution you'd like**
So, we can we define subaction using the `+` e.g.

`Action(....) + Action(....) ` so I am constantly doing more things to the object but I am just adding actions. Also, now actions are more composable, say if I wanna do the same subaction to two different objects I can just do `[Action(...), Action(...)] .+ Action(....)`

*Describe alternatives you've considered**
Hmm. None.

**Additional context**
Libraries in R like ggplot2 adopts this addition syntax to modify a plot. So I think it makes for Javis.jl too.

Wikunia:
Thanks for trying out Javis!

I'm quite confused to be honest 😄 
A `SubAction` is quite different to an Action. It basically defines the animated properties of the parental action.

The Action syntax doesn't allow to let an object appear it change color as an example. This can be only done using a SubAction. Examples can be found in tutorial no. 4

That being said I kind of like your idea of applying the same SubAction to two actions. It might be possible to have this in the syntax described in #100 

I would additionally like to have a way to change the frame range then I think.

If you don't mind: can you send me some of your Javis code that you tried such that I can see the problem more clearly and might offer a way how I would express this in Javis?

xiaodaigh:
> The Action syntax doesn't allow to let an object appear it change color as an example. This can be only done using a SubAction.

That's something I don't understand. Why is there such a distinction between the two? To me they are both just actions that can be applied to an object.

Wikunia:
An `Action` creates the object and has very simple possibilities like `Translation` and `Rotation` but for complicated stuff `SubAction` is needed. One might rename `Action` to `Object` and `SubAction` to `Action` if that makes more sense. 
Then one might want to remove those convenience functions like `Translation` and the action on the object needs to be defined inside what we currently call `SubAction`.

xiaodaigh:
> One might rename `Action` to `Object` and `SubAction` to `Action`

That would make much more sense! And that is my mental model anyway. Strip away any "action" from `Action` and let subaction control all the movements colors etc with `+` syntax!

Wikunia:
I'll leave this open such we don't loose track of it. I don't think we can do this in the next release. Unfortunately it's quite a breaking change. We should implement this quite soon as long as Javis is not used by many people.
[...]

PR #205, opened 2020-09-26
add julia_logo data and code

ric-cioffi:
added the julia logo example code (and coordinates) to reproduce the one on [discourse](https://discourse.julialang.org/t/fourier-draws-the-julia-logo/47309)

I didn't actually load the .gif yet because that's very heavy (>15MB) but you guys should be able to reproduce that as well.
Also, in the code I'm making use of both `OffsetArrays` and `FFTViews`, they are both not strictly necessary in the sense that the code can be refactored to remove them but it might take me some time to get the DFT indices right so I left it there, but I also didn't feel like adding them as dependencies just for an example.

Let me know if we want to change/add something :) 

## PR Checklist

**Link to relevant issue(s)**
Closes #


examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
...
+function draw_logo(n::Int; frames = 2160)
+    # TODO: (1) Fix indices so that I don't need to go to end-1 (a problem for very small n)
+
+    base_size = 7.5e-5
+    n_rotations = 2         # how many times the "tip" goes around the drawing
+
+    remap_idx(i::Int) = (-1)^(i - 1) * floor(Int, (i + 1) / 2)
+    remap_inv(n::Int) = 2n * sign(n) - 1 * (n > 0)
+
+    n0, nT = -Int((n - 1) / 2), Int((n - 1) / 2)
```
Wikunia, line 86: fails for odd `n`

examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
...
+    x = d * (1 - α) / cos(ϑ)                    # distance b/w p1 and triangle vertices
+
+    v1 = p1 + s * x * Point(exp((θ + ϑ) * im))    # first triangle vertex
+    v2 = p1 + s * x * Point(exp((θ - ϑ) * im))    # second triangle vertex
+
+    line(p1, p2, :stroke)                   # vector line
+    poly([p2, v1, v2, p2], :fill)           # vector tip
+end
+
+function draw_logo(n::Int; frames = 2160)
```
Wikunia, line 77: A brief explanation of what `n` is would be appreciated.

examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
...
+    javis(
+        vid,
+        [
+            BackgroundAction(1:frames, ground),
+            Action(
+                1:frames,
+                ids[0],
+                (args...) ->
+                    object(Point(base_size * cs[0]), size = abs(base_size * cs[1])),
+                Rotation(0.0, 0.0),
```
Wikunia, line 109: this is not needed, right?

examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
...
+                1:frames,
+                :tip,
+                (args...) -> object(
+                    Point(base_size * cs[indices[end - 1]]),
+                    "yellow",
+                    size = 4,
+                    style = :fill,
+                    opacity = 1.0,
+                ),
+                Rotation(0.0, 0.0, ids[indices[end - 1]]),
```
Wikunia, line 153: feels a bit odd to have a non existing rotation here but I don't really have a better way to express this :smile:

examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
...
+# Import the discrete fourier transform of our points and recast them in their original complex form
+ps = readdlm("julia_logo.csv", ',')
+cs = complex.(ps[:, 1], ps[:, 2]) |> FFTView
+
+vid = Video(800, 500)
+
+function ground(args...)
+    background("black")
+    sethue("black")
+    circle(Point(0, 0), 200, :fill)
```
Wikunia, line 14: feels odd to draw a black circle on a black background or in general having black as the default color.

Wikunia (commented):
Thanks again for the amazing work! This is just a very early view of the code with some comments that might be helpful or just questions where I lack knowledge :smile: 
I should have a look at FFT again to add some actual value.

Regarding the gif: Maybe you can upload it to YouTube if you don't mind and we can include the video of it in our documentation.

examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
+using Javis
+using DelimitedFiles
+using OffsetArrays, FFTViews # (not strictly necessary, but make DFT indexing more intuitive)
+
+# Import the discrete fourier transform of our points and recast them in their original complex form
+ps = readdlm("julia_logo.csv", ',')
```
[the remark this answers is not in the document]
ric-cioffi: I can definitely share the code, but it's 200+ LOC and requires a bunch of extra packages. Should I just add an explanation on how it was created?

examples/julia_logo.jl
```diff
@@ -0,0 +1,159 @@
...
+    push!(path, pos)
+    draw_line.(path[2:end], path[1:(end - 1)]; color = color)
+end
+
+import Javis.Point
+Point(c::Complex) = Point(real(c), imag(c))
+distance(p1::Point, p2::Point) = sqrt((p2.x - p1.x)^2 + (p2.y - p1.y)^2)
+
+# Draw a vector between points p1 and p2
+function vector(p1, p2; color = "white", opacity = 1 / 2, linewidth = 1, tipsize = 0.1)
```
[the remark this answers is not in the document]
ric-cioffi: Cool, didn't even know it existed. I actually spent an excessive amount of time getting the arrow tip right so I'm more than happy to replace it

examples/julia_logo.jl
ric-cioffi: Ah, you're right. I can add a check that makes sure n is always odd (not sure how things would change with an even number of points, but can check that too)

examples/julia_logo.jl
ric-cioffi: Yes, was left there from old code

examples/julia_logo.jl
ric-cioffi: Yes, I just needed the "tip" to track the last point but didn't know how else to do it

examples/julia_logo.jl
ric-cioffi: No reason, was also left from old code. I just thought yellow on black background was nicer, but we can change that as well

ric-cioffi:
Thanks for all the comments, I'll try to push the updates by tomorrow

examples/julia_logo.jl
Wikunia: I like the yellow one black 👌

examples/julia_logo.jl
Wikunia: It's basically a translation, right? But yeah this works fine.

examples/julia_logo.jl
Wikunia: Yeah or ceil the division maybe

examples/julia_logo.jl
ric-cioffi: Ah, I haven't worked with translation yet. Should I just use `Translation(p)` where `p` is the point I want to track? Because it does not really have `from` and `to` points

examples/julia_logo.jl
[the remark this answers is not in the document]
Wikunia: I read a short explanation of how you did it on Zulip but yeah would be great to have one here. Many point is that it's hard to get what is saved in the file itself. Like is it just points? Or is it some fourier calculation or ... 😂 I don't think code is necessary as it's not something about javis but people who are interested should be able to write the code by reading the explanation and of course some general understanding and further internet research just a brief overview of the steps one needs to do 😊

examples/julia_logo.jl
Wikunia: I think `Translation(p,p)` should do it but yeah it's not more readable. We actually think about this in #179 like defining the movement only inside the SubAction such that the current action is more like an object description which is static instead of having a movement itself.

examples/julia_logo.jl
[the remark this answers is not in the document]
ric-cioffi: Mmmh, one thing I could do is actually upload the .csv with the (already ordered) list of coordinates and then do the DFT in the example code itself. That would make the code easier to understand but would also need the `FFTW` package. Then I can add an explaination of how the set of ordered coordinates was created in the first place which, as you say, it's not something about javis but that people can look up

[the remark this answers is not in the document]
Wikunia: Yes that sounds like a good approach. Don't worry about dependencies. As it's "only" in the examples section we don't need to add it to dependencies or test dependicies of Javis

examples/julia_logo.jl
ric-cioffi: Alternatively I could define a convenience function `Track(p)` that does exactly the same thing (but that wouldn't really change things)

examples/julia_logo.jl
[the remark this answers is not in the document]
ric-cioffi: Excellent!

examples/julia_logo.jl
Wikunia: Oh don't worry about this. It was more a not for myself to think about ways to make this stuff either. I like your hack it's just that I'm a bit embarrassed that one needs a hack for this 😂

ric-cioffi:
I've implemented the suggested changes. The only one I could not use was the `arrow` function as it gives the following error:
```
ERROR: can't draw arrow between two identical points
```
not sure if the problem is that at any moment the points start all in the same place or is the fourier series itself that has two identical ponts (that'd be weird but who knows).
I have also uploaded a .txt file in the gifs with the YouTube link, feel free to move it around or use it differently.

Let me know if there's anything else that needs changing :)

Wikunia:
Great thanks for the updates! I'll have a look at the arrow function again. Looks great so far.

Wikunia:
@TheCedarPrince maybe you want to have a look as well and you can also add the video to the examples section with a link to the code as for follow path.
@ric-cioffi I assume you allowed us to add commits to this PR, right?

ric-cioffi:
> @ric-cioffi I assume you allowed us to add commits to this PR, right?

Yes

TheCedarPrince (commented):
This looks great! I only left one comment about if you wanted to use Luxor.arrow. I updated the examples section and readme with your example, @ric-cioffi.

@Wikunia feel free to check this out again with the changes I added.

examples/julia_logo.jl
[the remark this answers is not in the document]
ric-cioffi: Not really. It gives an error for trying to connect two identical points. I'm not sure if it is due to an initialization problem or if it's because at any point during the animation two points actually coincide (which might actually happen given how the fourier series are constructed)

Wikunia:
Thanks for the update @TheCedarPrince 

@ric-cioffi one can use:
```julia
isapprox(p1, p2) && return
sethue(color)
setopacity(opacity)
arrow(p1, p2; linewidth=linewidth)
return
```

but I like your version. Maybe we just add a comment that one can use `arrow`. 

@TheCedarPrince do you mind to add something like "Kindly provided by @ric-cioffi" in the examples section? And we can remove the YouTube link in the text file or maybe have it in the `.jl` file at the top for a more high-res rendering.

Wikunia:
Do you mind to have a look at my changes @ric-cioffi ? 
If that's okay I'll merge it to master

ric-cioffi:
Looks good to me :)

PR #206, opened 2020-09-29
Draw latex

Wikunia:
## PR Checklist

If you are contributing to `Javis.jl`, please make sure you are able to check off each item on this list:

- [x] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [x] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [x] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [x] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [x] Did I properly add test dependencies to the `test` directory (if applicable)?
- [x] Did I check relevant tutorials that may be affected by changes in this PR?
- [x] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**
Closes #170 

**How did you address these issues with this PR? What methods did you use?**
I compute the width and height of the latex string and then use it as a clipping region and a circle grows from the top left corner to make it appear like the text is drawn from left to right. This is quite similar to what I have done for #161

PR #224, opened 2020-10-10
Drawing LaTeX Improvements

TheCedarPrince:
## PR Checklist

**Link to relevant issue(s)**
#161 

The draw LaTeX PR was good (#206 ) but I noticed that the drawing animation seems to "swoop" into the frame. This is a result of using the `circle` object to cover up/uncover the LaTeX being animated. I propose to add additional shapes (such as rectangles) to modify this behavior.


TheCedarPrince:
@Wikunia, I noticed that there seems to be a strange behavior in my code in providing a rectangular glide. Here is my code:

```julia
function animate_latex(text, pos::Point, t, action)
    svg = get_latex_svg(text)
    action == :stroke && (action = :fill)
    if t >= 1
        translate(pos)
        pathsvg(svg)
        do_action(action)
        translate(-pos)
        return
    end

    pathsvg(svg)
    polygon = pathtopoly()
    w, h = polywh(polygon)

    translate(pos)
    pathsvg(svg)
    do_action(:clip)
    rect(O, t * w, h, :fill)
    translate(-pos)
end
```

And here is the output:
latex

Somehow, the `h` "height" return from `polywh` does not completely cover the LaTeX. Any thoughts as to why this is? Is this a bug in `polywh`?

Wikunia:
It gives you the width and height but not the starting point. You assume the starting point is the origin but this seems to be wrong in this case. You should be able to see it when drawing the rectangle in red and on top the latex.

TheCedarPrince:
Hey @Wikunia I tested this again and, though I changed the presumed point, I still have the same problem. Here is what the output of my modified code is now:

```julia
function animate_latex(text, pos::Point, t, action)
    svg = get_latex_svg(text)
    action == :stroke && (action = :fill)
    if t >= 1
        translate(pos)
        pathsvg(svg)
        do_action(action)
        translate(-pos)
        return
    end 
    
    pathsvg(svg)
    polygon = pathtopoly()
    w, h = polywh(polygon)
    sethue("red")
    rect(pos, t * w, h, :fill)
    pathsvg(svg)
    do_action(:clip)
    translate(-pos)
end
```

and output:

latex

It's almost like the LaTeX is being shifted somehow... Thoughts?

Wikunia:
I can't see anything to be honest. Why don't you draw the latex on top of the rectangle to see where it is exactly and why it doesn't match.

Wikunia:
Will you continue trying to tackle this @TheCedarPrince ?

TheCedarPrince:
Yes @Wikunia - it is on my radar to be worked on in v0.3.x. So, should be one of the first ones I tackle.

PR #225, opened 2020-10-11
Add Require stuff

### LINKED ISSUE #193: Conditional Loading of Packages

I have noticed recently that `using Javis` seems to be taking much longer than it used to. I ran some benchmarks and it was somewhat concerning:

**v0.1.5 Benchmarks:**

```julia
julia> @time using Javis
 41.913861 seconds (37.30 M allocations: 2.000 GiB, 3.49% gc time)
```

**v0.1.0 Benchmarks:**

```julia
julia> @time using Javis
[ Info: Precompiling Javis [78b212ba-a7f9-42d4-b726-60726080707e]
 21.673696 seconds (12.55 M allocations: 715.007 MiB, 0.83% gc time)
```

We have currently doubled the compile time of `Javis`! :scream: 

I would like to conditionally compile packages as they are needed. Obviously, some packages that are critical would always be loaded but I think we need to identify where the compile time issues come from.

Thanks to a suggestion from @mcabbott I think [`Requires.jl`](https://github.com/JuliaPackaging/Requires.jl) would help fix this problem.

Sov-trotter:
## PR Checklist

**Link to relevant issue(s)**
Closes #193 

Used the @requires method to load the javis_viewer.jl file since that's what uses Gtk
```
function __init__()
    @require Gtk="4c0ca9eb-093a-5379-98c5-f87ac0bbbf44" include("javis_viewer.jl")
end

```
Not sure if this is 100% correct but would be nice point to start from. I'd appreciate if someone could checkout this branch and provide feeback. :)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ LightXML = "9c8b4983-aa76-5018-a973-4c85ecc9e179"
 Luxor = "ae8d54c2-7ccd-5906-9d76-62fc9837b5bc"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 VideoIO = "d6d074c3-1acf-5d4c-9a43-ef38773959a2"
 
 [compat]
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -1,11 +1,13 @@
 module Javis
 
+using Requires 
+
 using Animations
 using Cairo: CairoImageSurface, image
 using ColorTypes: ARGB32
 using FFMPEG
-using Gtk
-using GtkReactive
+# using Gtk
+# using GtkReactive
 using Images
 using LaTeXStrings
 using LightXML
@@ -91,7 +93,7 @@ include("backgrounds.jl")
 include("svg2luxor.jl")
 include("morphs.jl")
 include("subaction_animations.jl")
-include("javis_viewer.jl")
+# include("javis_viewer.jl")
 include("latex.jl")
 include("transition2transformation.jl")
 include("symbol_values.jl")
@@ -425,5 +427,9 @@ export scaleto
 
 # custom override of luxor extensions
 export setline, setopacity, fontsize, get_fontsize, scale, text
+
+function __init__()
+    @require Gtk="4c0ca9eb-093a-5379-98c5-f87ac0bbbf44" include("javis_viewer.jl")
+end
 
 end


Wikunia:
Currently only on mobile. Does this work locally on a test script? Seems like it's failing in CI as the file doesn't get included properly. 
I'm still puzzler of how this should work 🤣 How should Julia know that it has to include the file when we call the javis viewer?

Sov-trotter:
Oh yeah. I am digging in a bit more. Just wanted the changes to be visible to everyone so that we know if it' the right direction.

Wikunia:
Thanks for having a look 😊 Maybe it's as simple as having `using Gtk` inside the `if` in which we call the image viewer?

Sov-trotter:
Okay. But we can extend it as:
-  The javis_viewer.jl has the code solely for the image viewer that can be made conditional based on whether Gtk is loaded or not. So I am trying something in towards that. 

Like you said this feature will be more beneficial for developers, Gtk.jl will be already precompiled in the first install of Javis, whereas it's the image viewer code that will be compiled everytime.

Sov-trotter:
@TheCedarPrince what do you think?

TheCedarPrince:
So after looking some more at Requires.jl, I am not sure if it actually does what I was expecting it to. I continue getting this error whenever I attempt to use liveview with an example. Here is my code:

```julia
   using Javis
   using LaTeXStrings

   function latex_ground(args...)
        translate(-200, -100)
        background("white")
        sethue("black")
        fontsize(50)
    end 
    
    video = Video(400, 200)
    javis(
        video,
        [
            BackgroundAction(1:1, latex_ground),
            Action((args...) -> latex(L"\mathcal{O}(\log{n})")), # default fontsize 50
            BackgroundAction(1:35, latex_ground),
            Action(
                (args...) -> latex(L"\mathcal{O}(\log{n})");
                subactions = [ 
                    SubAction(1:15, appear(:draw_text)),
                    SubAction(21:35, disappear(:draw_text)),
                ],
            ), # default fontsize 50
        ],
        pathname = "latex.gif",
        liveview = true
    )   
```

and here is the error

```julia
julia> include("view.jl")
[ Info: Precompiling Javis [78b212ba-a7f9-42d4-b726-60726080707e]
ERROR: LoadError: UndefVarError: _javis_viewer not defined
Stacktrace:
 [1] javis(::Video, ::Array{Action,1}; framerate::Int64, pathname::String, liveview::Bool, tempdirectory::String) at /home/src/Projects/javis/src/Javis.jl:217
 [2] top-level scope at /home/src/Projects/javis/view.jl:12
 [3] include(::String) at ./client.jl:457
 [4] top-level scope at REPL[2]:1
in expression starting at /home/src/Projects/javis/view.jl:12
```

@Sov-trotter , you are free to continue exploring Requires.jl but @Wikunia 's simple solution of a conditional may be the best case here. I was hoping for something more elegant which is why I was attracted to Requires.jl. 

Let me know if you have any more questions!

Sov-trotter:
Okay. Now I see. Require will be helpful where things are not as convoluted/inter-dependent like here in Javis. I also tried some benchmarks and using Javis took around 25s for me. Closing this for now.  :)

Sov-trotter:
I was digging around a bit more and came across https://github.com/JuliaReinforcementLearning/GridWorlds.jl/commit/f46e78c3ff224039adb42a3362edee5c8a053428

Sov-trotter:
Not sure if it's possible but just putting it down here.
@Wikunia do you think we can solve this by creating a `Task` at the global level which imports `Gtk` and schedule it conditionally on the basis of `liveviewer` flag?
https://docs.julialang.org/en/v1/base/parallel/

Wikunia:
I think we should try this again with v1.6 of Julia and decide whether it's really necessary. Additionally it might make more sense to split up the package into some parts.

PR #228, opened 2020-10-15
Implementation of render and act 

Wikunia:
## For reviewers:
- This is build on top of #226 so that should be reviewed/merged first.

If you are contributing to `Javis.jl`, please make sure you are able to check off each item on this list:

- [ ] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [ ] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [ ] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [ ] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [ ] Did I properly add test dependencies to the `test` directory (if applicable)?
- [ ] Did I check relevant tutorials that may be affected by changes in this PR?
- [x] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**
Closes #227 

**How did you address these issues with this PR? What methods did you use?**
Instead of using `javis(video, objects)` one now has to define all objects beforehand and run
`render(video)`. This removes the weird array syntax. 

Additionally instead of using `+` to add an action to an object the function is called `act!` now.

An example looks like this:

```julia
video = Video(500, 500)

translate_action = Action(Translation(100, 10))
BackgroundObject(1:100, ground)
c1 = Object((args...) -> circle(O, 25, :fill))
c2 = Object(30:70, (args...) -> circle(O, 25, :fill), Point(0, 40))
act!(c1, translate_action)
act!(c2, translate_action)

render(video, pathname = "current/_trans.gif")
```

PR #226, opened 2020-10-11
first step of Action -> Object, SubAction -> Action

Wikunia:
## PR Checklist

If you are contributing to `Javis.jl`, please make sure you are able to check off each item on this list:

- [ ] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [x] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [x] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [x] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [x] Did I properly add test dependencies to the `test` directory (if applicable)?
- [ ] Did I check relevant tutorials that may be affected by changes in this PR?
- [x] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**
Closes #179 
Closes #204 

**How did you address these issues with this PR? What methods did you use?**
Renaming from Action -> Object and SubAction -> Action
Adding an action to an object can be done with `+` (this will later be changed to `act!`)

```julia
using Javis

function ground(args...)
    background("black")
    sethue("white")
end

video = Video(800, 600)
javis(
    video,
    [
        BackgroundObject(1:100, ground),
        Object(1:100, (args...) -> circle(O, 50, :fill), Point(100, 10)) +
        Action(1:100, appear(:fade)),
    ];
    pathname = "current/object.gif",
)
```

PR #230, opened 2020-10-15
Working with the variable instead of the id

### LINKED ISSUE #229: Use the variable instead of the symbol to access position, angle, etc
With the new syntax in #228 it doesn't make sense to have a symbol attached to an object.
It would be much easier to use the variable name of the object itself.

As an example: `red_ball = Object(1:100, circ)` instead of `Object(1:100, :red_ball, circ)`

Wikunia:
## For reviewers:
- This is build on top of #226 and #228 so these should be reviewed/merged first.

**Link to relevant issue(s)**
Closes #229

The object stores the result now (in a vector to be mutable) and can be used itself in `Transitions` and for `pos`, `angle` etc.

Example of usage:
```
demo = Video(500, 500)

BackgroundObject(1:100, ground)
start_scale = Object((args...) -> (1.0, 1.0))
circle_obj = Object((args...) -> circ())
act!(circle_obj, Action(1:25, Scaling((1.0, 1.5))))
act!(circle_obj, Action(Rel(25), Scaling((2.0, 1.0))))
act!(circle_obj, Action(Rel(25), Scaling(start_scale)))
act!(circle_obj, Action(Rel(25), Scaling(2.0)))
Object((args...) -> circ(Point(-100, 0)))
```


TheCedarPrince (approved):
LGTM - I can see no issues with this. I will have to further examine if there are any memory leaks occurring while using Javis as we saw, but I don't think this is an immediate problem. Furthermore, let's do a think about if we should have `act!` return the mutated object as I thought we could do it here. Wasn't the point of this PR, but I was happy to know why this doesn't work.

Either way, great work!

PR #236, opened 2020-10-17
Lazy loading for Gtk

### LINKED ISSUE #193: Conditional Loading of Packages

Sov-trotter:
## PR Checklist

**Link to relevant issue(s)**
Closes #193 

Continuation of https://github.com/Wikunia/Javis.jl/pull/225


TheCedarPrince:
Hey @Sov-trotter - thanks for continuing to look at this. I am still getting the same error that I got last time being this along with some warnings about overwriting:

```julia
(@v1.5) pkg> activate .
 Activating environment at `/home/src/Projects/javis/Project.toml`

julia> include("view.jl")
[ Info: Precompiling Javis [78b212ba-a7f9-42d4-b726-60726080707e]
WARNING: Method definition __init__() in module Javis at /home/src/Projects/javis/src/Javis.jl:101 overwritten at /home/src/Projects/javis/src/Javis.jl:441.
  ** incremental compilation may be fatally broken for this module **

ERROR: LoadError: UndefVarError: _javis_viewer not defined
Stacktrace:
 [1] javis(::Video, ::Array{Action,1}; framerate::Int64, pathname::String, liveview::Bool, tempdirectory::String) at /home/src/Projects/javis/src/Javis.jl:227
 [2] top-level scope at /home/src/Projects/javis/view.jl:94
 [3] include(::String) at ./client.jl:457
 [4] top-level scope at REPL[2]:1
in expression starting at /home/src/Projects/javis/view.jl:94
```

Here is the code I was using:

```julia
using Javis
using Random

function ground(args...)
    background("white")
    sethue("black")
end

function draw_line(p1 = O, p2 = O, color = "black", action = :stroke, edge = "solid")
    sethue(color)
    setdash(edge)
    line(p1, p2, action)
end

function circ(p = O, color = "black", action = :fill, radius = 25, edge = "solid")
    sethue(color)
    setdash(edge)
    circle(p, radius, action)
end

function info_box(video, action, frame)
    fontsize(12)
    box(140, -210, 170, 40, :stroke)
    text("10-20 EEG Array Readings", 140, -220, valign = :middle, halign = :center)
    text("t = $(frame)s", 140, -200, valign = :middle, halign = :center)
end

function electrode(
    p = O,
    fill_color = "white",
    outline_color = "black",
    action = :fill,
    radius = 25,
    circ_text = "",
)
    sethue(fill_color)
    circle(p, radius, :fill)
    sethue(outline_color)
    circle(p, radius, :stroke)
    text(circ_text, p, valign = :middle, halign = :center)
end

electrode_locations = [
    O,
    Point(-70, 0),
    Point(70, 0),
    Point(-140, 0),
    Point(140, 0),
    Point(0, 70),
    Point(-50, 70),
    Point(50, 70),
    Point(0, -70),
    Point(-50, -70),
    Point(50, -70),
    Point(115, -80),
    Point(-115, -80),
    Point(115, 80),
    Point(-115, 80),
    Point(40, -135),
    Point(-40, -135),
    Point(-190, -10),
    Point(190, -10),
    Point(-40, 135),
    Point(40, 135),
]

electrode_names = [
    "Cz",
    "C3",
    "C4",
    "T3",
    "T4",
    "Pz",
    "P3",
    "P4",
    "Fz",
    "F3",
    "F4",
    "F8",
    "F7",
    "T6",
    "T5",
    "Fp2",
    "Fp1",
    "A1",
    "A2",
    "O1",
    "O2",
]

radius = 15
indicators = ["tomato", "darkolivegreen1", "gold1", "white"]
demo = Video(500, 500)
javis(
    demo,
    [
        BackgroundAction(1:10, ground),
        Action(
            :inside_circle,
            (args...) -> circ(O, "black", :stroke, 140, "longdashed"),
        ),
        Action(:head, (args...) -> circ(O, "black", :stroke, 170)),
        Action(
            :vert_line,
            (args...) ->
                draw_line(Point(0, -170), Point(0, 170), "black", :stroke, "longdashed"),
        ),
        Action(
            :horiz_line,
            (args...) ->
                draw_line(Point(-170, 0), Point(170, 0), "black", :stroke, "longdashed"),
        ),
        Action(
            :electrodes,
            (args...) ->
                electrode.(
                    electrode_locations,
                    rand(indicators, length(electrode_locations)),
                    "black",
                    :fill,
                    radius,
                    electrode_names,
                ),
        ),
        Action(:info, info_box),
    ],
    pathname = "eeg.gif",
    framerate = 1,
    liveview = true
)
```

It seems like you are getting closer based on the grid worlds issue you found here: https://github.com/JuliaReinforcementLearning/GridWorlds.jl/commit/f46e78c3ff224039adb42a3362edee5c8a053428 Maybe ask @findmyway for some help? He used Javis to make the logo for JuliaReinforcementLearning. :smile:

Sov-trotter:
Yeah. I am digging in a bit more since we are now clear about what we want and also that we have a use case. I'll let you know once I make sure that things work well on my end. :)))

findmyway:
Thanks for @TheCedarPrince 's invitation.

Hi @Sov-trotter ,

I notice that you've commented out some tests. I guess some important parts are missed.

To keep tests working as usual, you need to add the `GTK` and `GtkReactive` in the `Extras` and `Targets` section of the root Project.tom. 
Just like this:

https://github.com/JuliaReinforcementLearning/ReinforcementLearningEnvironments.jl/blob/0932006ceafae3073ac569021bf925200d91e1cf/Project.toml#L22-L32

Another way is to create a new environment inside of tests folder. (By creating a Project.toml file in the tests) . But I forget when is this feature enabled. (I guess since Julia@v1.3)

Another tip.

Given that GtkReactive relies on GTK, you can simply require `GtkRequires` and then 

```julia
using .GtkReactive
using .GtkReactive.GTK
```

So that one can simply enable those extra features by `using GtkReactive` instead of using two packages.
(I haven't test it. But it should work ;)

Sov-trotter:
Thank you so much @findmyway for the help. I was able to get things working with the examples @TheCedarPrince quoted but the tests were failing for some reason. For now I fixed it by using `GtkReactive.Gtk` qualification for Gtk methods.

Also there are two warnings viz.
```
WARNING: using GtkReactive.label in module Javis conflicts with an existing identifier.
WARNING: using GtkReactive.textbox in module Javis conflicts with an existing identifier.
```
I am not really sure what can be done on this side but I'd love to work on any suggestions. :)

Wikunia:
Thanks @Sov-trotter and @findmyway 

The warnings appear because both GtkReactive and Javis export `label` and `textbox`. This can be fixed by using `import` instead of `using`. 

I think you can remove the test extras from the Project.toml as you have them in `test/Project.toml` 

Additionally I am wondering whether you deleted the reference images.

Sov-trotter:
Oh. Sorry for that, since the tests were successful for the first time on my machine, I though those were supposed to be ignored.
Replacing `import` for `using` does remove the warnings but it creates issues with `Requires`,
```
 Warning: Error requiring GtkReactive from Javis:
│ LoadError: LoadError: LoadError: UndefVarError: @guarded not defined
```

Wikunia:
You probably want to use something like:
```
import Luxor
import Luxor: Point, @layer
```
in this case `import Gtk: @guarded` and probably some others. (maybe GtkReactive :smile:  It's not my part of the code base so my knowledge is limited)

Refs are the reference images which stay constant such that the newly computed images are checked against them.

Wikunia:
Great work. Now is it somehow possible that the user doesn't have to write `using GtkReactive` ? 
I would like that it's automatically loaded when the user sets the keyword `liveview=true`

Sov-trotter:
Yeah was thinking of the same. That side would be kinda different from the requires route I guess. One way I am thinking of is where the true flag is checked for? Not sure though

Wikunia:
Maybe you can double check that this works:
```julia
 if liveview == true
        @eval import .GtkReactive
        @eval import .GtkReactive: slider, signal
        @eval using .GtkReactive.Gtk
        _javis_viewer(video, length(frames), actions)
        return "Live preview started."
    end
```
(inside `src/Javis.jl`)

and then remove the `import` from `javis_viewer.jl`

Sov-trotter:
Oh cool. What I wanted to try was
```
if liveview == true
        include("javis_viewer.jl")
        _javis_viewer(video, length(frames), actions)
        return "Live preview started."
    end
``` 
while the file would remain the same, when I call the function with `liveviewer= true`, I get a `ERROR: could not open file /home/ezio/.julia/dev/Javis/javis_viewer.jl`

Wikunia:
Yours might fail because it tries to load it in the current working directory instead of the javis src directory?

Unfortunately mine doesn't work. 
The problem with loading GtkReactive as a user might be that we introduce the warnings again so the user would need to write import?

Feels a bit weird to write that as the user 😕 
Hope we can figure it out

Sov-trotter:
>The problem with loading GtkReactive as a user might be that we introduce the warnings again so the user would need to write import?

Sorry you mean with Requires or the if liveview=true condition method?

Wikunia:
Sorry I mean if the user has to write:
```
using Javis
using GtkReactive
```
like it is currently the case with this PR.

Sov-trotter:
Ah yeah. But I don't think that will bring up the warnings since we have it fixed inhere https://github.com/Wikunia/Javis.jl/blob/f5c299aabf8ab6dcf23b0698c9199344ef35bf12/src/javis_viewer.jl#L1-L3. The main role of `using GtkReactive` as a user would be to invoke the Requires API?

Wikunia:
I think there is a problem if the user writes `label` in his/her code, right?

Sov-trotter:
Sorry didn't get you?

Wikunia:
Sorry my explanation is probably kinda lacking :smile: 
If the user has
```julia
using Javis
using GtkReactive

function drawing(args...)
  label() # should be Javis but is ambiguous as GtkReactive also exports it, right?
end
```

I hope this makes clear what I'm worried about. Anyway best would be if the user doesn't need to write `using GtkReactive`

Sov-trotter:
Oh. I think the warning can be avoided by using a qualification ? But yeah this seems to take us down to more such future warnings. :/

Wikunia:
It can be avoided by using `import` I think but yeah it's not a nice user experience. Based on our discussion that mostly @TheCedarPrince and myself are experiencing the slow down because of precompilation it's not really ideal. I hope we can finde a way around this.

TheCedarPrince:
Hey @Sov-trotter - I posted about this on the Julia Discourse and it seems like we may have a few potential answers (see here: https://discourse.julialang.org/t/dynamic-code-loading-to-use-packages-on-demand/48655). Hopefully, one of these can address this problem.

TheCedarPrince:
Hey @Sov-trotter do you still want to work on this or should I take it over? Thanks!

Sov-trotter:
Yeah. I'd love to. Sorry I am a bit slow due to some other commitments. I'll get it done soon. :)

Wikunia:
No hurry @Sov-trotter We probably need at least 10 more days to bring v0.3 and I think this issue can't be added before that anyway as it needs a new dependency.

Sov-trotter:
So following up on the discussion over discourse, I am still getting the world age error. Not sure what's incorrect though. :/

TheCedarPrince:
Hey @Wikunia and @Sov-trotter - I checked this out again and I am really stuck on what to do with this issue. I am debating about closing it altogether sadly. :disappointed:

Wikunia:
I feel like it's not strictly necessary but appreciate the effort @Sov-trotter .
Would have be great but I think we have more important issues to tackle.
Feel free to close this @Sov-trotter if you don't want to further work on it.

Sov-trotter:
Okay. Sorry though I am not able to commit time rn(caught up in academic stuff currently), I think till mid-december. I think you can close it for now. We can however revisit this later as the package grows.

PR #242, opened 2020-10-20
Ability to disable an action after its last defined frame

### LINKED ISSUE #238: Ability to disable that an action gets called with latest value
Currently 
```julia
obj = Object(1:100, ...)
act!(obj, Action(1:10, anim, sethue()))
```

means that the color stays like the last value. This is sometimes not what the user wants or expects. 
It often it as should still be the default but one should be able to disable that functionality.

An example is
```julia
act!(obj, Action(1:10, anim1, sethue()))
act!(obj, Action(21:30, anim2, sethue()))
act!(obj, Action(11:20, anim3, sethue()))
```
In this case one would expect that `anim2` is the last but actually `anim3` will hold from `31:100`.  

It would be best to check if the same function is called already but that might be hard to achieve in some cases.

An easy fix is to have something like
```julia
act!(obj, Action(11:20, anim3, sethue(); keep=false))
```

Wikunia:
## For reviewers
This is done on top of #237 so that should be reviewed first.

**Link to relevant issue(s)**
Closes #238

There is a keyword for `Action` now called `keep` which defaults to true. 
One can disable this functionality with `; keep = false`. An example can be found in the issue #238


TheCedarPrince (approved):
LGTM - do you think it could be possible to also add this functionality to an `Object` as well? Feel free to merge regardless.

Wikunia:
Can you clarify what you mean by add this functionality to an Object? An object is always only present for the frames it is defined for.

TheCedarPrince:
Wait, this functionality I was thinking of is essentially the same as #247 . Nevermind for the moment! Thanks @Wikunia

PR #250, opened 2020-10-21
Translation -> anim_translate and co + unit tests

### LINKED ISSUE #245: Improve syntax for `Translation`, `Scaling`, and `Rotation`

Currently, the nomenclature is not very clear on what these are exactly and on the backend of Javis, uses a large amount of code to create structs to handle the behavior supported by each one of these functionalities. 

Modifying these functionalities from structs to actual functions is desired as it would make development and maintenance of Javis much easier in respect to these functionalities. 

@Wikunia and I explored ideas for different syntax and looked at the following ideas:

```julia
Action(1:100, [Point(1,100)], translate())
Action(1:100, anim, translate())
anim = Animation([0.0, 1.0], [O, Point(1,100)], [linear())]
Action(1:100, [Point(20, 20), Point(1,100)], translate())
```

We came to the conclusion that this was the best idea:

```julia
anim_translate(O, Point(1, 100))
anim_translate(O, Point(1, 100); easing=sineio())
Action(1:100, sineio(), anim_translate(1, 100))
```

### LINKED ISSUE #246: Better Way to Define Animation

Right now, defining an Animation to use for an animation is somewhat clunky as it inherits its format directly from the [`Animations.jl` package](https://github.com/jkrumbiegel/Animations.jl). This format I think is perfect for Animations.jl and those who have used it in the past but can be a bit difficult for a Javis user unfamiliar with Animations.jl.

Here is an example of the Javis syntax which invokes an `Animation` object:

```julia
anim = Animation([0.0, 1.0], [O, Point(1,100)], [linear()])
```

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #245

`anim_X` is now converted to a struct which holds the information `from` and `to`. It can be either a point, value, scale whatever or an `Object`. The interpolation is done on the range from `0.0` to `1.0` with Animations.jl. When actually getting the value like a point for `translate` there is a new step now: `interpolation_to_transition_val` which takes the interpolation value and converts it to a transition value like `Point` or `Scale`. 

It probably needs more tests and docstring changes but docstrings are also changed in #237 so that should be revviewed first anyway.


Wikunia:
The functionality should be working now. Thorough tests with more test cases would be good. I haven't changed docstrings yet. 
Will do so later. You can review this already for functionality and syntax changes if you want @TheCedarPrince and if everything makes sense I can add docstrings or do you want me to add docstrings before?

src/Javis.jl
```diff
@@ -373,13 +366,14 @@ end
...
-export Line, Translation, Rotation, Transformation, Scaling
-export val, pos, ang, get_value, get_position, get_angle
+export Line, Transformation
+export val, pos, ang, scl, get_value, get_position, get_angle, get_scale
 export projection, morph_to
 export appear, disappear, rotate_around, follow_path, change
 export rev
 export scaleto
 export act!
+export anim_translate, anim_rotate, anim_rotate_around, anim_scale
```
TheCedarPrince, line 376: I am not a huge fan of the names `anim_rotate` and `anim_rotate_around`. I finally figured out what we could name these two things:

- `anim_rotate` stays `anim_rotate`
- `anim_rotate_around` could instead become `anim_revolve` where "revolve" is defined as rotation around a distant point.

src/structs/Action.jl
```diff
@@ -44,17 +41,12 @@ It can be defined in many different ways.
     - **Default:** The default is `linear()`
     - Possible simple easing functions is `sineio()` for more check
         [Animations.jl](https://jkrumbiegel.github.io/Animations.jl/stable/)
-- func is either a `Function` or a `Transition`
-    - This is the actual action that is applied to the parent object.
-    - It can be either a general function which takes in the following four arguments
+- func is the function that describes the actual action
+    - It can be a general function which takes in the following four arguments
         - video, object, action, rel_frame
```
TheCedarPrince, line 46: Perhaps provide better explanation into what each of these four arguments are (e.g. where they come from, how they are used by the `Action`, etc.)

src/structs/Scale.jl
```diff
@@ -0,0 +1,11 @@
...
+    x::Float64
+    y::Float64
+end
+
+Base.:+(s1::Scale, s2::Scale) = Scale(s1.x + s2.x, s1.y + s2.y)
+Base.:-(s1::Scale, s2::Scale) = Scale(s1.x - s2.x, s1.y - s2.y)
+Base.:*(s1::Scale, f::Number) = f * s1
+Base.:*(f::Number, s1::Scale) = Scale(f * s1.x, f * s1.y)
+Base.:*(s1::Scale, s2::Scale) = Scale(s1.x * s2.x, s1.y * s2.y)
+Base.:/(s1::Scale, s2::Scale) = Scale(s1.x / s2.x, s1.y / s2.y)
```
TheCedarPrince, line 11: Curious about this - wasn't it discussed on the Julia Slack that dispatching these base Julia definitions is a sort of an anti-pattern? Could you explain why you are providing this syntactic ability?

TheCedarPrince (changes requested):
A few syntax suggestions and questions in my comments. A few more comments here:

1. You wanted me to see if I could create more test cases for `anim` functions. What do you think is lacking currently in the test suite?

2. It is nice to see the LOC decreasing! Happy that we are reducing the amount of code having to be maintained.

src/Javis.jl
Wikunia: I like that `anim_rotate` will suggest `anim_rotate_around` in your editor. `revolve` would be much harder to find. One would probably need to do `?anim_rotate`. I personally don't find it that intuitive either (maybe because I'm not a native speaker)

src/structs/Action.jl
Wikunia: Can this be done in the first tutorial? I think `rel_frame` can be explained but the other three are quite intuitive or not? One should have defined a `Video` an `object` and an `action` at this point already.

src/structs/Scale.jl
Wikunia: Can you link me to that discussion if it still exists? :smile: 
I think it makes sense to define `+` and `-` and `*` such that interpolation works. It's not type piracy as we own `Scale` so I think it should be fine?

src/Javis.jl
TheCedarPrince: Hmmm... I agree with what you mean and the TAB suggestion is very convenient to have. I was borrowing the syntax idea of `revolve` from 3D modeling tools where that sort of syntax is commonly used. Let's keep the syntax the way it is.

src/structs/Scale.jl
TheCedarPrince: This was the part of the discussion I was referencing off the Slack (wasn't lost yet!):

image

I agree with your logic behind it not being type piracy. I just wanted to make sure we aren't introducing bad practices into Javis via this approach. Feel free to resolve this issue if you feel things are not a problem.

src/structs/Action.jl
TheCedarPrince: That's true. I will add this to issue #254 so we don't forget the point made here.

src/structs/Scale.jl
Wikunia: Yeah I think `Object + Action` would fall into that category of syntactic sugar but this is like
```julia-repl
julia> using Luxor
[ Info: Precompiling Luxor [ae8d54c2-7ccd-5906-9d76-62fc9837b5bc]

julia> Point(1,2)*2
Point(2.0, 4.0)
```

TheCedarPrince (approved):

Issue #214, opened 2020-09-30
Better Warnings for Objects without Background


TheCedarPrince:
**Is your feature request related to a problem? Please explain.**

It is hard to sometimes keep track of an Action being defined outside of the frames for a given animation. Especially in the case of defined BackgroundActions. Without a BackgroundAction, the Action will not render properly (if at all) and there ought to be  warning a user to this possible problem.

**Describe the solution you'd like**

I would like to add the ability to alert a user that a BackgroundAction frame is not present for a given Action that is defined outside the range provided by the BackgroundAction. A user could potentially define multiple different backgrounds so the solution will have to account for this.

Wikunia:
Similar to #213 is this a question? Would also move this to v0.2.x

Wikunia:
Seems like we fixed this already.

PR #263, opened 2020-10-30
preprocess_frames! function

### LINKED ISSUE #260: Preprocessing of Frames for `get_javis_frame`

Currently, it is difficult to return a single frame of an animation to a user in an easy and performant manner.

Javis should provide functionality with the `get_javis_frame` function to preprocess a chosen frame and deliver it back to the user quickly. @Wikunia, I am not sure exactly how we would go about implementing this but you thought this should be relatively easy/straightforward to fix.

**Additional context**
Linked to issue #92 for allowing a user to easily get the frame they want.

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #260 

I've created a `preprocess_frames!` function which computes the correct frames for all objects and actions. It can also be used for #214 at one point. The function can be used when neither the live preview should be shown nor a gif/mp4 should be rendered but instead a single frame should be computed with `get_javis_frame`.

Additionally I've updated the unit tests to use this functionality.


src/Javis.jl
```diff
@@ -119,6 +119,37 @@ function projection(p::Point, l::Line)
     return c * v + o
 end
 
+"""
+    preprocess_frames!(objects::Vector{<:AbstractObject})
+
+Computes the frames for each object and action based on the user defined frames that the
+user can provide like `Rel` and `:same`.
```
TheCedarPrince, line 126: Only thing here is to change `Rel` to match the `RFrame` syntax. Thanks!

TheCedarPrince, line 126: Could you add to the docstring about why this preprocess is needed for the `get_javis_frame` method? I understand that it defines CURRENT_OBJECT, but I had to do a bit of poking around to figure out what exactly it was doing.

TheCedarPrince (changes requested):
Only two docstring comments but otherwise, LGTM. Great work! Do you think this would be an appropriate time to also export both `get_javis_frame` and `preprocess_frames!`?

src/Javis.jl
Wikunia: I've added a docstring. `CURRENT_OBJECT` is not that important I think. It can maybe get removed there at some point.

TheCedarPrince (approved):

Issue #247, opened 2020-10-21
Make an `:all` Symbol for Frame Ranges


TheCedarPrince:
**Is your feature request related to a problem? Please explain.**

Right now, one can define an Object such that,

```julia
BackgroundObject(1:200, ground) # creates the background of the animation

Object(1:100, myfunc)
Action(:same, myaction)
Action(myaction) # the same as the preceding line

Object(100:200, myotherfunc)
Action(myotheraction) # as the unit range of 100:200
```

It would be nice to have a unit range that allows an action to be defined over the entirety of the animation.

**Describe the solution you'd like**

Here is an idea:

```julia
BackgroundObject(1:200, ground) # creates the background of the animation

Object(1:100, myfunc)
Action(:same, myaction)
Action(myaction) # the same as the preceding line

Object(100:200, myotherfunc)
Action(myotheraction) # as the unit range of 100:200
Action(:all, mybestfunc) # defined over the range of 1:200 frames
```

Furthermore, the `:all` keyword would allow an Action to be defined anywhere in the code and not immediately after a prior `Action`.

PR #265, opened 2020-10-30
Feature global frames

### LINKED ISSUE #235: Global frame definition in Action (v0.3)
It's sometimes a little hard to define the frames in a relative fashion inside an `Action` (v0.3).
If I want to define two objects that appear one after another but should disappear at the same global time one has to write some kind of weird formula :smile: 

We have the ability to use `Rel` already which I haven't really used at all yet :smile: and I think `Glob` makes more sense for this.
So instead of writing 
```julia
o = Object(101:200, ...)
act!(o, Action(10:20, ...))
```
one can write
```julia
o = Object(101:200, ...)
act!(o, Action(Glob(110:120), ...))
```

Wikunia:
## PR Checklist

 - We should have this in #264 and maybe find another tutorial where it makes sense.

**Link to relevant issue(s)**
Closes #235 

The global frame computation needs to know the parent object so `set_frames` needs that. 
I fixed a bug that `:same` is now `:same` as the action or object before. Before this it was actually `:same` as the object before.
We might want to have an `:all` keyword for this see #247 .

### DIFF
--- a/test/unit.jl
+++ b/test/unit.jl
@@ -49,7 +49,7 @@
         end
     end
 
-    @testset "RFramesative frames" begin
+    @testset "Relative frames" begin
         video = Video(500, 500)
         Object(1:100, (args...) -> 1)
         # dummy object doesn't need a real function
@@ -59,7 +59,7 @@
         rm(test_file)
     end
 
-    @testset "GFramesal frames" begin
+    @testset "Global frames" begin
         video = Video(500, 500)
         Object(1:100, (args...) -> 1)
         # dummy object doesn't need a real function
@@ -143,7 +143,7 @@
         end
     end
 
-    @testset "RFramesative frames" begin
+    @testset "Relative frames" begin
         video = Video(500, 500)
         o1 = Object(1:100, (args...) -> 1)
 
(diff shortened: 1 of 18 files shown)


test/unit.jl
```diff
@@ -132,11 +143,11 @@
         end
     end
 
-    @testset "Relative frames" begin
+    @testset "RFramesative frames" begin
```
TheCedarPrince: Should this line by `RFrames`?

test/unit.jl
```diff
@@ -49,16 +49,27 @@
...
```
TheCedarPrince: Should this line by `RFrames`?

test/unit.jl
```diff
@@ -49,16 +49,27 @@
...
         Object(1:100, (args...) -> 1)
         # dummy object doesn't need a real function
-        object = Object(Rel(10), (args...) -> 1)
+        object = Object(RFrames(10), (args...) -> 1)
         test_file = render(video)
         @test Javis.get_frames(object) == 101:110
         rm(test_file)
     end
 
+    @testset "GFramesal frames" begin
```
TheCedarPrince: Should this line by `GFrames`?

TheCedarPrince (approved):
LGTM - I think you fixed all my comments while I was streaming so I appreciate your fixes!

PR #273, opened 2020-11-03
Warning if no background

### LINKED ISSUE #214: Better Warnings for Objects without Background

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #214 

The video now saves a range for which a background is defined. The `preprocess_frames!` function then checks whether some frames don't have a background and throws a warning. 
Additionally I've changed some test cases to not throw a warning :smile:


Wikunia:
I have no idea why format is failing. I've run the `format` command. Maybe you can double check this @TheCedarPrince

src/Javis.jl
```diff
@@ -144,6 +147,10 @@ function preprocess_frames!(objects::Vector{<:AbstractObject})
         append!(frames, collect(get_frames(object)))
     end
     frames = unique(frames)
+    if !(frames ⊆ CURRENT_VIDEO[1].background_frames)
+        @warn("Some of the frames don't have a background: In this case: ",
```
TheCedarPrince, line 151: "background:" to "background."

src/Javis.jl
```diff
@@ -144,6 +147,10 @@ function preprocess_frames!(objects::Vector{<:AbstractObject})
...
+             setdiff(frames, CURRENT_VIDEO[1].background_frames))
+    end
```
TheCedarPrince, line 153: I feel that the warning is a bit verbose:

```
┌ Warning: Some of the frames don't have a background: In this case:
│   setdiff(frames, (CURRENT_VIDEO[1]).background_frames) =
│    1-element Array{Int64,1}:
│     11
└ @ Javis /home/src/Projects/javis/src/Javis.jl:151
```

Perhaps instead we could do this: 

```julia
        @warn("Some of the frames don't have a background: In this case: $(
             setdiff(frames, CURRENT_VIDEO[1].background_frames))")
```

which I think produces the nicer warning:

```
┌ Warning: Some of the frames don't have a background: In this case: [11]
└ @ Javis /home/src/Projects/javis/src/Javis.jl:151
```

TheCedarPrince (changes requested):
This makes sense - I found one typo and made a minor suggestion on implementation. Only other question that I had is should we seek to identify which Object does not have a background defined?

Wikunia:
> Only other question that I had is should we seek to identify which Object does not have a background defined?
I don't think we can.

TheCedarPrince (approved):

PR #276, opened 2020-11-07
Docstrings for v0.3

Wikunia:
## PR Checklist

  - I've realized that we missed to bound `Hungarian`...

Docstrings for the `anim_` functions. I've added a `# Possibilities` to show the different argument possibilities this time. Hope that makes sense.
I've also added a docstring for `act!` as that was missing. Maybe it can be simplified as the possibilities are quite simple.


src/structs/Transitions.jl
```diff
@@ -19,9 +36,45 @@ struct Rotation{T<:Real} <: AbstractTransition
...
+    anim_rotate_around
+
+Animate the rotation of the attached object (see [`act!`](@ref)) around a point.
+Similiar function: [`anim_rotate`](@ref) to rotate or spin an object
+
+# Example
+```julia
+Background(1:100, ground)
+obj = Object((args...) -> rect(O, 50, 50, :fill), Point(100, 0))
+act!(obj, Action(1:50, anim_rotate_around(2π)))
```
TheCedarPrince, line 70: Is this option defined in the Options section here? I don't clearly see it.

src/structs/Transitions.jl
```diff
@@ -37,5 +90,22 @@ Scaling(fs::Tuple, ts) = Scaling(Scale(fs...), ts)
...
+# Example
+```julia
+Background(1:100, ground)
+obj = Object((args...) -> rect(O, 50, 50, :fill), Point(100, 0))
+act!(obj, Action(1:50, anim_scale(1.5)))
+```
+
+# Options
+- `anim_scale(ts)` scales from the current scale to `ts`.
+- `anim_scale(fs, ts)` scales from `fs` to `ts`.
```
TheCedarPrince, line 108: I get that `ts` and `fs` represent "to scale" and "from scale" but could you add a type definition of whether they accept Floats, Ints, Reals? Thanks!

TheCedarPrince (changes requested):
I made a few grammar changes and typo fixes as well as cleaned up some of the docstrings. Just a few comments to resolve and then I think it is great.

src/structs/Transitions.jl
Wikunia: There are quite a lot of different ways like Symbol, Number and object

src/structs/Transitions.jl
Wikunia: Which option?

Wikunia: I forgot defining a position there. Copy and paste error :smile:  
Thanks

src/structs/Transitions.jl
Wikunia: Maybe it's clearer now. I didn't document the symbol there as it's more for internal usage.

TheCedarPrince (approved):
LGTM! Feel free to merge.

Issue #249, opened 2020-10-21
Make Better Error Messages for v0.3.0


TheCedarPrince:
**Is your feature request related to a problem? Please explain.**

In Javis, it has often been unclear to know where an error occurs via the `Action` and `SubAction` syntax. This should not be! Therefore, let's do an exhaustive check of the new syntax we are making for Javis v0.3.0 to assist with better debugging. A check list of things to review could be:

- [x] Check `Object` syntax by defining bad functions or bad frame ranges or providing inappropriate `id` symbols.
- [x] Check `Action` syntax be providing bad frame ranges, modify the wrong `Object`, put it out of range of a previously defined `Object`, etc. #278 
- [x] Use invalid background functions for `Background`, attempt to modify `Background` with an `Action`, provide wrong unit ranges, provide unit ranges that do not cover all of the animation (#214), etc.
- [x] Put multiple video sources for render, have render show nothing even though `Object` items are defined, etc.

**Describe the solution you'd like**

Better warning messages for these items in case a user is having difficulty with producing a solution.

TheCedarPrince:
Can we close this @Wikunia ?

Wikunia:
I don't think we have done any of these tests. At least I haven't.

Wikunia:
Unfortunately I find the points very vague i.e What do you mean by: invalid background functions? Like the function doesn't have `args...` as an input? 
What do you mean by "multiple video sources"

TheCedarPrince:
I think we pretty thoroughly addressed this with #273 and #278 . Should be good now!

PR #278, opened 2020-11-09
Warning if action is defined out of object frames

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
#249 point 2

I check the length of the object that exists and compare whether the action is defined for a range outside of that frames.
The warning message tries to give the best clue possible by giving the number of the action and object and the frame range that it could be in vs the one it is in.

### DIFF
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -139,8 +139,8 @@ Shows a warning if some frames don't have a background.
 function preprocess_frames!(objects::Vector{<:AbstractObject})
     compute_frames!(objects)
 
-    for object in objects
-        compute_frames!(object.actions; parent = object)
+    for (i, object) in enumerate(objects)
+        compute_frames!(object.actions; parent = object, parent_counter=i)
     end
 
     # get all frames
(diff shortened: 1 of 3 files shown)


TheCedarPrince:
@Wikunia I love this!!! This is so useful! However, there seems to be a strange behavior that I noticed:

```julia
video = Video(100, 100)
Background(1:20, (args...) -> 1)
act!(Object(1:11, (args...) -> 2), Action(1:20, anim_translate(0, 10)))
act!(Object(1:11, (args...) -> 2), Action(1:30, anim_translate(0, 10)))
render(video; liveview = true)
```

Where the warning message becomes:

```
┌ Warning: An Action defines frames outside the range of the parental object.
│             In particular Action #1 for Object #2 is defined for frames
│             1:20 but the object exists only for 11 frames.
│             (Info: the Background is counted as Object #1)
└ @ Javis /home/src/Projects/javis/src/util.jl:31
┌ Warning: An Action defines frames outside the range of the parental object.
│             In particular Action #1 for Object #3 is defined for frames
│             1:30 but the object exists only for 11 frames.
│             (Info: the Background is counted as Object #1)
└ @ Javis /home/src/Projects/javis/src/util.jl:31
"Live preview started."
```

It seems like the Action incrementer is not increasing? Thoughts?

Wikunia:
It's the first action of each object, right? It's seems easier to navigate this per object instead of looking at all actions.

TheCedarPrince (approved):
LGTM - merging now.

PR #283, opened 2020-11-11
seperate LaTeXSVG into latexsvgfile.jl 

ArbitRandomUser:
## PR Checklist

**Link to relevant issue(s)**
Closes #


Wikunia:
Make sure to run:
```julia
using JuliaFormatter
format(".")
```
in your root directory.

src/latexsvgfile.jl
```diff
@@ -0,0 +1,22 @@
+#file with the LaTeXSVG constant
+export LaTeXSVG
```
Wikunia, line 2: Is there a reason you want to export this?

ArbitRandomUser: i thought variables had to be exported ; i guess thats only for modules ?
does include("filename") work like #include in  c?

ArbitRandomUser: docstring for include says , `Evaluate the contents of the input source file in the global scope of the containing module. Every module (except those defined with baremodule) has its own definition of include, which
  evaluates the file in that module.`
so export isnt needed i believe . Should i remove export ?

Wikunia: export would mean that users of Javis can access it without using `Javis.` in front of it. `include` should be working fine without `export`. I would therefore remove it.

Wikunia:
Thanks @ArbitRandomUser

PR #295, opened 2020-11-26
Change can now also set value

### LINKED ISSUE #279: Have a set function similar to change which just sets the value
Currently `change` can only be used when having values that define a `-` method and `*` to have interpolation.
Sometimes it might make sense to change other arguments like string though. Basically just setting the argument to a new value.

I would like to have:

```julia
Action(1:1, set(:value, "new value"))
```
or something similar. 

It would only work with actions that have a single frame.
Therefore, we might want to allow `Action(1, set(:value, "new value"))` instead of forcing a unit range.

Another option would be have:

```julia
Action(1:1, change(:value, "new value"))
```

so we don't define a new function name but only have a single value instead of a pair. It will be set at the last frame defined such that it doesn't break for 

```julia
Action(1:2, change(:value, "new value"))
```

and just sets it for the frames `1:2` and then can be used with: `keep = false`

```julia
Action(1:2, change(:value, "new value"); keep=false)
```

to fall back to the default after frame 2.

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #279 

I've added a `change` method which accepts everything but pairs and just set the keyword specified to that value. I additionally found a bug that appeared with `; keep=false` as this didn't work for `change`. It now resets the dictionary every time such that it works now.


TheCedarPrince (approved):
LGTM :smiley:

PR #307, opened 2020-12-22
Latex $$ fix

### LINKED ISSUE #306: [BUG] latex() assumes $$ in LaTeXString
**Describe the bug**

`get_latex_svg()` assumes the LaTeXString passed is bracketed by `$$`. However, not all LaTeXString constructors auto append the `$$`. From the LaTeXString [README](https://github.com/stevengj/LaTeXStrings.jl):

> Finally,` you can use the lowest-level constructor LaTeXString(s). The only advantage of this is that it does not automatically put $ at the beginning and end of the string.

**To Reproduce**

v1.5.3
Mac & Linux
3. Javis version (i.e output of `] status Javis` in the REPL)
v0.3.3

```
using Javis, LaTeXStrings
Javis.get_latex_svg(LaTeXString(L"\sqrt{x}") == Javis.get_latex_svg(LaTeXString("\\sqrt{x}")
```

agerlach:
## PR Checklist

**Link to relevant issue(s)**
Closes #306

Check if `$` in the LaTeXString before removing.

### DIFF
--- a/src/latex.jl
+++ b/src/latex.jl
@@ -92,7 +92,10 @@ function get_latex_svg(text::LaTeXString)
         svg = LaTeXSVG[text]
     else
         # remove the $
-        ts = text.s[2:(end - 1)]
+        ts = text.s
+        if ts[1] == '$'
+            ts = ts[2:(end - 1)]
+        end
         command = `tex2svg $ts`
         try
             svg = read(command, String)
--- a/src/latex.jl
+++ b/src/latex.jl
@@ -86,6 +86,14 @@ function animate_latex(text, pos::Point, t, object)
     translate(-pos)
 end
 
+"""
+    strip_eq(text::LaTeXString)
+
+Strips `\$\$` from `text.s` if present and returns the resulting string. 
+
+# Arguments
+- `text::LaTeXString`: a LaTeX string
+"""
 function strip_eq(text::LaTeXString)
     ts = text.s
     if ts[1] == '$'
--- a/src/latex.jl
+++ b/src/latex.jl
@@ -94,6 +94,7 @@ function strip_eq(text::LaTeXString)
     ts
 end
 
+# \todo update LaTeXSVG cache to use output of strip_eq as the key. See https://github.com/Wikunia/Javis.jl/pull/307#issuecomment-749616375
 function get_latex_svg(text::LaTeXString)
     # check if it's cached
     if haskey(LaTeXSVG, text)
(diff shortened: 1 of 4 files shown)


Wikunia:
Thanks a lot! Do you mind to change the version directly in the Project.toml? :wink:

Wikunia:
Regarding the test failure:
You can only use LaTeX commands which are in the `LaTeXSVG` dict

agerlach:
> Regarding the test failure:
> You can only use LaTeX commands which are in the `LaTeXSVG` dict

This doesn't make sense to me by looking at the source. Tests pass locally. The CI logs seem to indicate that `tex2svg` is not installed.

agerlach:
Are you saying it needs to be included in the `Dict` constructor in `latexsvgfile.jl`? If, so it will end up just bypassing the code that needs tested. I can understand why in your other tests you want to bypass `txt2svg` and use a reference svg, but does that makes sense here?

Wikunia:
Good point we might want to save the `ts` in the dict and not the text, right? Then you can do your step before checking if it is cached. It currently fails because it's not cached and then we try to render it but haven't installed the npm module on GitHub.

agerlach:
That would work, but will require changing the types used in LaTeXSVG as the key won't be `LaTeXString` anymore. I'm unclear on what the upstream affects of that would be. 

Two other thoughts:

1. Why not just install the npm module?  
2. I could bypass all this by moving to code to a function, e.g `strip_eq(txt::LaTeXString) = txt[1] == '$' ? txt[2:(end-1)] : txt`. The test will then just be a test of this function.

Wikunia:
I like the first option but I have not enough knowledge how to do it. I'm fine with option 2 for now

agerlach:
OK, I will do 2. However, as a future note, from a cache perspective it would be better to cache based on the stripped string because now you can have two different keys that produce the same svg., e.g. `\\sqrt{x}` and `$\\sqrt{x}$`.

Wikunia:
Totally agree. Maybe add a `Todo:` in the code 😊

agerlach:
done

src/latex.jl
```diff
@@ -86,13 +86,21 @@ function animate_latex(text, pos::Point, t, object)
     translate(-pos)
 end
```
Wikunia, line 88: If you don't mind a short docstring would be great :smile:

agerlach: done

Wikunia:
Thanks a lot! Enjoy the holidays :wink:

PR #309, opened 2020-12-26
add example projection, fix other examples

matbesancon:
Adding example with a rotating circle and its projection onto the positive orthant.

Based on:
https://matbesancon.github.io/post/2020-12-24-chains_sets2/

I also fixed the other examples which did not accommodate running them from another folder, relative paths should use `@__DIR__`.

Also, running Javis often leads to an error when ImageIO and ImageMagick are not installed, I added these


matbesancon:
weird, I did run the formatter with the blue code style on the only added file

examples/follow_path.jl
```diff
@@ -2,12 +2,12 @@ using Javis, Animations
 
 function ground(args...)
     background("white")
-    sethue("black")
+    return sethue("black")
```
Wikunia, line 5: Why did you add `return` here?

Wikunia:
Thanks for the example!
Hmm seems like the formatter isn't happy yet. Did you run `using JuliaFormatter; format(".")` in the root folder?
Do you mind adding your example to https://github.com/Wikunia/Javis.jl/blob/master/docs/src/examples.md ?

Wikunia:
@TheCedarPrince we should figure out why ImageIO and ImageMagick were needed for @matbesancon .

matbesancon:
>  Hmm seems like the formatter isn't happy yet. Did you run using JuliaFormatter; format(".") in the root folder?

I ran it on the examples only. Let me check if I run on root

matbesancon:
I just ran JuliaFormatter on everything, it seems some files were not compliant

matbesancon:
Here is the result I get by formatting the file the action complains about:
image

matbesancon:
@Wikunia there is a problem with the javis_viewer.jl file, basically the Formatter can run multiple times and always finds something to change

Wikunia:
I made a PR for your fork @matbesancon . When I rerun formatter locally it says that it's already formatted. It behaved weird here such that I changed the comment a bit (removed newlines)

matbesancon:
Yep just saw and merged it thanks

examples/follow_path.jl
Wikunia: I prefer `sethue("black")` without the `return` as it's not needed afterwars.

matbesancon: it was added by the formatter

matbesancon: if you don't want to set returns at the bottom of all functions, is should be specified in the Formatter TOML I believe

matbesancon: also, see:
https://github.com/invenia/BlueStyle#method-definitions
it is advised by the blue style, that's why it's there by default

Wikunia:
Interesting as it wasn't there before your change

matbesancon:
did you try running formatter on everything? That's when it happened

Wikunia:
I always run it on root. I think I'll change the config for it. First we (@TheCedarPrince and I) should figure out why InageMagick is needed.

Wikunia:
Alright. That took much longer than hoped but looks good to me. I'll merge this and change the return in an upcoming PR where I'll fix the configuration for the formatter as well.

PR #311, opened 2021-01-07
set always_use_return to false in formatter

Wikunia:
## PR Checklist

I'm not a fan of `always_use_return` therefore I disabled it. :wink: I think it makes sense to explicitly use `return` when one wishes to use the value afterwards which isn't the case for things like `sethue`.


TheCedarPrince (approved):
LGTM and is reasonable! Also noticed you added the ImageIO as a dependency in the CHANGELOG. Was that to address the error that @matbesancon reported?

Wikunia:
Yes it's used by Images.save with png 
I think we both had it installed locally that's why it worked. Not sure why it passed the CI tests though

PR #318, opened 2021-03-13
Added new line

XRFXLP:
Added new line to separate the segments.

## PR Checklist

### DIFF
--- a/docs/src/tutorials/tutorial_1.md
+++ b/docs/src/tutorials/tutorial_1.md
@@ -64,7 +64,7 @@ end
 ```
 
 > **NOTE:** One may wonder, "why is `args...` needed in the `ground` function?"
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/tutorial_1.md
```diff
@@ -64,7 +64,7 @@ end
 ```
 
 > **NOTE:** One may wonder, "why is `args...` needed in the `ground` function?"
-> Each user-defined function gets three arguments `video`, `object`, and `frame` provided by Javis
+> Each user-defined function gets three arguments `video`, `object`, and `frame` provided by Javis
```
Wikunia, line 67: I think I would also add a `.` at the end of the sentence.

XRFXLP: Thanks, I've added that now.

Issue #277, opened 2020-11-09
Latex alignment


Wikunia:
**Is your feature request related to a problem? Please explain.**
It's currently not possible to use `halign` and `valign` for `latex` whereas they are options for `text`. 

**Describe the solution you'd like**
In general `latex` should provide the same options as `text`.

TheCedarPrince:
I think this issue could be tied to #224 - happy to take a look/work on it some.

Wikunia:
Yeah might use similar functions

codejaeger:
@Wikunia @TheCedarPrince can I try working on this issue?

Wikunia:
Sure. Let us know if you need some help.

PR #321, opened 2021-03-20
Add alignment options to latex text display

codejaeger:
## PR Checklist

**Link to relevant issue(s)**
Attempt to solve issue #277

Presently 3 test cases fail which is due to the different width and height parameters used from the previous version in `latex.jl.` Due to this the corresponding frames get generated differently. This approach does not require generation of a polygon from the path created by the latex svg. @Wikunia @TheCedarPrince any suggestions?

Here is an example:

```julia
using Javis
using LaTeXStrings

function ground(args...)
    background("white")
    sethue("black")
end

function draw_latex(video, action, frame)
    black_red = blend(O, Point(0, 150), "black", "red")
    setblend(black_red)
    fontsize(50)
    latex(
        L"""\begin{equation}
        \left[\begin{array}{cc} 
        2 & 3 \\  4 & \sqrt{5} \\
        \end{array} \right] 
        \end{equation}""",
        O,
        valign = :middle,
        halign = :center
    )
end

demo = Video(300, 200)
Background(1:1, ground)
Object(draw_latex)
render(demo; pathname = "latex.gif", framerate=1)
```

Output:

latex1


src/latex.jl
```diff
@@ -3,8 +3,11 @@
...
 latex(text::LaTeXString) = latex(text, O)
 latex(text::LaTeXString, pos::Point) = latex(text, pos, :stroke)
+latex(text::LaTeXString, pos::Point, valign::Symbol, halign::Symbol) =
+    latex(text, pos, :stroke, valign = valign, halign = halign)
 latex(text::LaTeXString, x, y) = latex(text, Point(x, y), :stroke)
-
+latex(text::LaTeXString, x, y, valign::Symbol, halign::Symbol) =
+    latex(text, Point(x, y), :stroke, valign = valign, halign = halign)
 """
     latex(text::LaTeXString, pos::Point, object::Symbol)
```
Wikunia, line 12: this is missing the updated function signature.
Needs also explanation of the arguments later and the difference in positional and keyword arguments there as you have defined them above.

src/latex.jl
```diff
@@ -74,16 +83,34 @@ function animate_latex(text, pos::Point, t, object)
...
     end
 
-    pathsvg(svg)
-    polygon = pathtopoly()
-    w, h = polywh(polygon)
+    w, h = svgwh(svg)
+
+    halignment = findfirst(isequal(halign), [:left, :center, :right, :centre])
+
+    # if unspecified or wrong, default to left, also treat UK spelling centre as center
```
Wikunia, line 90: I think I would throw at least a warning when the value is wrong.

src/latex.jl
```diff
@@ -74,16 +83,34 @@ function animate_latex(text, pos::Point, t, object)
...
+    textpointx = pos.x - [0, w / 2, w][halignment]
+
+    valignment = findfirst(isequal(valign), [:top, :middle, :bottom])
+
+    # if unspecified or wrong, default to baseline
+    if valignment === nothing
+        valignment = 3
+    end
+
+    textpointy = pos.y - [0, h / 2, h][valignment]
```
Wikunia, line 106: One might want to change this to `(0, h / 2, h)` to avoid alocations.

src/latex.jl
```diff
@@ -74,16 +83,34 @@ function animate_latex(text, pos::Point, t, object)
...
 
-    translate(pos)
+    translate(Point(textpointx, textpointy))
```
Wikunia, line 108: maybe use `point = Point(...)` to later write `-point`.

src/svg2luxor.jl
```diff
@@ -229,6 +229,22 @@ set_attr(::Val{:d}, args...) = nothing
 set_attr(::Val{:id}, args...) = nothing
 set_attr(t, args...) = @warn "No attr match for $t"
 
+"""
+    svgwh(svg)
+
+Return width and height of svg on screen.
```
Wikunia, line 235: I'm unsure of what the "on screen" means.

Wikunia (changes requested):
Thanks for the PR. I've made some mostly minor comments. Will have a look at the test cases now. Additionally make sure to add a test case for this functionality as well.

Wikunia:
@codejaeger I've updated the reference images. Make sure that when you add test cases for this functionality that you only use latex from the cached `LaTeXSVG` as otherwise CI tests will fail.

codejaeger:
Thanks for the review @Wikunia , I will make the changes accordingly.

src/latex.jl
```diff
@@ -9,7 +9,7 @@ latex(text::LaTeXString, x, y) = latex(text, Point(x, y), :stroke)
 latex(text::LaTeXString, x, y, valign::Symbol, halign::Symbol) =
     latex(text, Point(x, y), :stroke, valign = valign, halign = halign)
 """
-    latex(text::LaTeXString, pos::Point, object::Symbol)
+    latex(text::LaTeXString, pos::Point, object::Symbol, valign::Symbol, halign::Symbol)
```
Wikunia, line 12: they should be keyword arguments here as well.

src/latex.jl
```diff
@@ -23,10 +23,16 @@ Can be added to `Luxor.jl` graphics via [`Video`](@ref).
 - **The `latex` method must be called from within an [`Object`](@ref).**
 
 # Arguments
+Positional:
 - `text::LaTeXString`: a LaTeX string to render.
 - `pos::Point`: position of the upper left corner of the latex text. Default: `O`
     - can be written as `x, y` instead of `Point(x, y)`
 - `object::Symbol`: graphics objects defined by `Luxor.jl`. Default `:stroke`.
+Keyword:
+- `valign::Symbol`: vertical alignment with respect to the specified `pos` parameter
```
Wikunia, line 32: please mention the default value and move this below `Available objects:` as that describes the `object` parameter.

src/latex.jl
```diff
@@ -75,44 +81,43 @@ end
...
-        translate(pos)
-        pathsvg(svg)
-        do_action(object)
-        translate(-pos)
-        return
-    end
 
     w, h = svgwh(svg)
-
     halignment = findfirst(isequal(halign), [:left, :center, :right, :centre])
```
Wikunia, line 86: use a tuple structure as well :wink:

Wikunia (changes requested):
Thanks for the updates. As mentioned earlier when running the CI tests for `latex` functions make sure that you either add it to `LaTeXSVG` or use svgs that are cached already.
Would like to see test cases for `animate_latex` with different `valign` and `halign` as well

codejaeger:
> As mentioned earlier when running the CI tests for `latex` functions make sure that you either add it to `LaTeXSVG` or use svgs that are cached already.

Oh, I was a bit confused by what you meant by 'using cached latex' before. It is clear now. I hope the tests pass in the CI in the upcoming commit. Thanks!

src/latex.jl
Wikunia: not fixed yet.

codejaeger: Oh sorry, I thought I had made the change. Anyways, I pushed recently again to improve codecov and they are reflected in them. Thanks.

Wikunia: seems to be still wrong on GitHub, right?

src/latex.jl
```diff
@@ -20,6 +23,7 @@ Can be added to `Luxor.jl` graphics via [`Video`](@ref).
 - **The `latex` method must be called from within an [`Object`](@ref).**
 
 # Arguments
+Positional:
```
Wikunia, line 26: please follow the style guide for these docstrings as listed here:
https://github.com/invenia/BlueStyle#keyword-arguments

They can also be seen in some other methods like our `render` function

src/latex.jl
```diff
@@ -56,36 +66,59 @@ render(demo; pathname = "latex.gif")
...
 
-function animate_latex(text, pos::Point, t, object)
+function animate_latex(text, pos::Point, t, valign::Symbol, halign::Symbol, object)
     svg = get_latex_svg(text)
     object == :stroke && (object = :fill)
+
+    w, h = svgwh(svg)
+    halignment = findfirst(isequal(halign), (:left, :center, :right, :centre))
+
+    if halignment === nothing
```
Wikunia, line 89: We would like to cover every single line using our test cases.
You can test for warnings using `@test_logs`

codejaeger:
Thanks @Wikunia for your patient reviews! I will make the changes as requested.

codejaeger:
Since the functions required to previously get the width and height of the svg latex (e.g. `polywh`) are not used in the `animate_latex` function anymore those are not covered by any tests. Should I add separate tests for that @Wikunia ?

Wikunia:
If you don't mind a small test case where you create a simple polygon and measure the width and height would be nice.
Not sure whether we'll need the function in the future.

test/svg.jl
```diff
@@ -43,6 +43,42 @@ end
...
+    )
+    Object(
+        (args...) ->
+            Javis.animate_latex(L"8", O - Point(20, 20), 2, :bottom, :right, :stroke),
+    )
+    Object(
+        (args...) ->
+            Javis.animate_latex(L"8", O - Point(-20, 20), 0, :middle, :center, :stroke),
+    )
+    @test_logs (:warn,) (:warn,) render(video; tempdirectory = "images", pathname = "")
```
Wikunia, line 77: a small comment on why it throws a warning here would be great. Thanks again for your work!

TheCedarPrince:
Hey @codejaeger - thank you for working on this PR! I have been busy with work and relocation so I have been a bit silent. @Wikunia has been directing you with this PR so I am not going to interrupt but just wanted to say that all tests passed on my machine and that the work looks great! Thanks so much Jaeger!

Wikunia (approved):

Wikunia:
Can you add a small line to the `# Unreleased` line in the Changelog.md file for this?

codejaeger:
@Wikunia I updated the `Changelog`. Anything else?

Wikunia:
Thanks a lot @codejaeger

PR #323, opened 2021-03-26
[WIP]: Add dev mode for faster animation

### LINKED ISSUE #166: Javis Animation Dev Mode

When developing Javis animations that take hundreds of frames, I have found that when I run the `javis` command it has to recompile and make all the hundreds of frames. I am impatient and do not want to wait when developing! 

Therefore, I think we could implement a `kwarg` like `debug::Bool` or `dev_mode::Bool` in the `javis` function that will do the following:

1. When `dev_mode == True` activate development mode
2. Compile all frames currently defined in the `javis` function and save them as image files in a temp directory
3. Finish compiling frames to filetype.

Upon running of the `javis` function with `dev_mode == True` again with more frames added:

1. If development mode is still active, then check what images have been rendered
2. Skip frames that were previously rendered 
3. Save new frames as image files in previous temp directory
4. Combine images together to final filetype. 

Therefore, saving time on development compilation by having frames that were previously rendered be saved.

**Additional context**

Based on discussion with @Wikunia !

codejaeger:
## PR Checklist

**Link to relevant issue(s)**
Attempt to close #166

The idea is to automatically track the updated frames in case the `dev_mode` is specified while creating the videos. This definitely has its limitations and cannot always track every unchanged frame i.e. if the image of the frame is left unchanged but somehow that frame was referred to or used since last render it will still get created in the next render.

Currently, tests need to be added to benchmark the improvements (if any). @Wikunia @TheCedarPrince could you give a initial review if possible?
Thanks!


Wikunia:
Will have a look later this weekend. Probably Sunday. Do you mind giving some code to go into dev mode and a bit more explanation to what should happen?
Thanks in advance m testing with morphing might fail based on your explanation so far.

codejaeger:
Yes sure @Wikunia.
```
using Javis

function ground(args...)
    sethue("black") # pen color
end

myvideo2 = Video(500, 500; dev_mode=true)
Background(1:70, ground)
red_ball = Object(1:70, (args...)->object(O, "red"), Point(100,0))
render(myvideo; pathname="tutorial_1.gif", compute_frames=:latest)
```
This is the way one should use it. I am still on the benchmarking part, trying to see what is the outcome. I did not find a test so far where not re-drawing the frames had any significant speed up. I will try to find one and add it as a test in a subsequent commit as test. Thanks.

TheCedarPrince:
Sure @codejaeger - I could probably do a cursory review today to give some feedback. 😄

Wikunia:
Thanks! What's the `:latest` part? Can one specify the frames one want to recompute? And does this work for gif and mp4?

codejaeger:
`:latest` is to signify to re-compute/draw only the frames which are thought to be different. If not specified, it should work just as it was before. I tried it with `gif`, I will do the same with `mp4` and let you know. No, currently there is no way to take in the frame range to compute as input. I will add that after this. This seemed to me a simpler way to provide a speed up.

Sov-trotter:
@codejaeger, Sorry for poking my nose in here, but I totally understand the `Image.load()` latency issue here as I had faced something similar while trying to reduce the lag in Jupyter Viewer. One approach you could try is replace `frame_image = Images.load("$(tempdirectory)/$(lpad(filecounter, 10, "0")).png")`  with something "in-memory" eg, like fetch that particular frame from an array of frames(stored previously). Again that might lead to increased allocations but it's worth giving a try.

Wikunia:
Lol we had the same thought @Sov-trotter just in a different issue #166

codejaeger:
Will try that @Sov-trotter, thanks for the suggestion.

codejaeger:
@Wikunia I tried the approach of storing the frames in memory. In a crude way, there is a performance speed up of 25% when I try with examples generating 300 and 500 frames. I will add tests for these. I wanted to ask does this speed up seem reasonable against the increased memory consumption?

Wikunia:
I think it makes sense to provide the option with bigger memory consumption but for faster rerendering.

PR #334, opened 2021-05-20
Flexible extension to RFrames

### LINKED ISSUE #331: Extension to RFrames
When specifying a relative frame range it would be would be nice to have an option to specify if the reference frame was the previous object's first or last frame. A particular use case is when objects need to be added to the canvas one after another based on their starting frame.

Add the argument `from_start::Bool` to `RFrames` to indicate the frame range for this object. 
Example:
```
ball = Object(Rframes(5:10, true), (args...)->somefunc())
```

A workaround is to create a function to access the starting frame of the desired object in a way like `start(obj)`. The only problem seems to be when the previous object (or a reference to it) is not known.

**Additional context**
Both the methods were suggested by @Wikunia on Javis' zulip [thread](https://julialang.zulipchat.com/#narrow/stream/253420-javis/topic/Extension.20to.20Rframes.3F).

codejaeger:
## PR Checklist

**Link to relevant issue(s)**
Closes #331 

As discussed in the issue there were 2 approaches to make `RFrames` more flexible. One was to supply additional flags to accommodate the different cases. Other is to let the user specify the starting reference and (optionally) the end frame. 
The latter allows a wide range of options like specifying the frame range based on a previously created object.
Any suggestions @Wikunia @TheCedarPrince @Sov-trotter ?

### DIFF
--- a/test/frames.jl
+++ b/test/frames.jl
@@ -1,0 +1,74 @@
+@testset "Check RFrame shorthands" begin
+    function dummy(args...) end
+    video = Video(10, 10)
+    Background(1:20, dummy)
+    a = Object(1:3, (args...) -> O) # 1:3
+    b = Object(RFrames(1:3), (args...) -> O) # 4:6
+    c = Object(RFrames(1:3, prev_start()), (args...) -> O) # 5:7
+    d = Object(RFrames(prev_start(), prev_last()), (args...) -> O) # 5:7
+    render(video; tempdirectory = "images", pathname = "")
+    @test a.frames.frames == 1:3
+    @test b.frames.frames == 4:6
+    @test c.frames.frames == 5:7
+    @test d.frames.frames == 5:7
+end
+
+@testset "Check RFrame helper function for previous object" begin
+    function dummy(args...) end
+    video = Video(10, 10)
+    Background(1:20, dummy)
+    a = Object(1:3, (args...) -> O) # 1:3
+    b = Object(RFrames(2:5, prev_start(), default_last()), (args...) -> O) # 3:6
+    c = Object(RFrames(2:5, prev_last()), (args...) -> O) # 8:11
+    d = Object(RFrames(0:3, prev_start(c)), (args...) -> O) # 8:11
+    e = Object(RFrames(2:5, prev_last(b)), (args...) -> O) # 8:11
+    render(video; tempdirectory = "images", pathname = "")
+    @test a.frames.frames == 1:3
+    @test b.frames.frames == 3:6
+    @test c.frames.frames == 8:11
+    @test d.frames.frames == 8:11
+    @test e.frames.frames == 8:11
+end
+
+@testset "Check RFrame helper function for previous action" begin
+    function dummy(args...) end
+    video = Video(10, 10)
+    Background(1:20, dummy)
+    a = Object(3:15, (args...) -> O) # 3:15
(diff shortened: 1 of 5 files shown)


test/frames.jl
```diff
@@ -0,0 +1,74 @@
...
+    @test c.frames.frames == 5:7
+    @test d.frames.frames == 5:7
+end
+
+@testset "Check RFrame helper function for previous object" begin
+    function dummy(args...) end
+    video = Video(10, 10)
+    Background(1:20, dummy)
+    a = Object(1:3, (args...) -> O) # 1:3
+    b = Object(RFrames(2:5, prev_start(), default_last()), (args...) -> O) # 3:6
```
Wikunia, line 21: I find this a little hard to read. What exactly is the meaning of `default_last()` when combined with `prev_start` here?

Wikunia:
Thanks for opening this PR @codejaeger 
I find some of the variants currently a bit hard to read. Have we thought of the following before?
Another idea might be to use `Frames(prev_start()+5, prev_start()+10)` to provide a start and end frame. Or maybe easier
`Frames(prev_start()+5, 10)` and the second parameter is always the length. I would more or less leave `RFrames` as it is then and just have a generic approach with `Frames` where the user can do whatever. Default to length of the object action seems reasonable to me but using `Frames(prev_start(); stop=100)` would make it straightforward I think.

Wikunia:
I think a good way to have a lot of flexibility would be to use:
```julia
macro Frames(start, len)
         return :(()->$start:$len)
end
```
and then with simple functions like this:
```julia
prev_start() = CURRENT_OBJECT[1].frames.frames[1]
```
This would be possible: 
```julia
@Frames(prev_start(), 10)
```
Javis just needs to call the anonymous macro function then to get the frames.
We have to check how we get something like `prev_start` work with actions as well and have maybe an `parent_start` to get the `object` instead of the previous action as an example.

Additionally maybe we can provide a `macro` with `start, stop` instead of `start, len` if that is desired.

Wikunia:
@codejaeger you can let me know here if you want to continue working on it. If not I might implement this myself in the next couple of days.

codejaeger:
Hey @Wikunia , here are my views and thoughts on the approach. Your idea seems a much more readable way to do the job than what the PR proposes. I have just a few doubts on the approach.
1. Is the usage going to be similar to `Object(@Frames(prev_start(), 10), (args...) -> dummy())`?
2. The `stop` parameter seems a global frame number to me. The way frames are stored in actions is relative to the Object. So this parameter does not seem very useful for action but it is important for Objects anyways.
3. The reason I was able to adjust `parent_start` to actions was because I had the type of the elem the frame belonged to available from inside the `get_frames` method.
4. I tried this code and right now it seems the CURRENT_OBJECT array is not updated at the end of Object creation. Is that a bug? I don't see it being used anywhere so it might have gone undetected.

Wikunia:
1. Yes it will be that syntax
2. We might want to have it relative to either the global one or the parent object. Similar to how the frames work now for the action
3. I'll need to look into this more
4. The current object only changes inside render as it should be independent of when it was created.
5. I think it should work well with actions and should just be relative to the parent object.

Wikunia:
Can we close this @codejaeger ?

codejaeger:
Yes, sure @Wikunia .

PR #337, opened 2021-05-26
[GSoC'21]Livestreaming support

### LINKED ISSUE #91: Supporting live streaming like Manim

Sov-trotter:
## PR Checklist

**Link to relevant issue(s)**
Closes #91 

Streaming is done via ffmpeg. I have used `@async` for the streaming to happen asynchronously and free up the julia shell,
For now the streaming can be terminated by manually killing the ffmpeg process.


Sov-trotter:
To view the livestream on can use `ffplay udp://0.0.0.0:8080` or add the address as input to a media source in OBS.

src/Javis.jl
```diff
@@ -183,6 +184,7 @@ Renders all previously defined [`Object`](@ref) drawings to the user-defined `Vi
 - `tempdirectory::String`: The folder where each frame is stored
     Defaults to a temporary directory when not set
 - `liveview::Bool`: Causes a live image viewer to appear to assist with animation development
+- `liveview::Bool`: Livestream the rendered animation to the local network
```
Wikunia, line 187: should be `livestream`

src/Javis.jl
```diff
@@ -262,6 +265,14 @@ function render(
         @error "Currently, only gif and mp4 creation is supported. Not a $ext."
     end
 
+    if livestream == true
```
Wikunia, line 268: could be `if livestream` actually we should do that for `if liveview` above too :smile:

Wikunia (commented):
Just two minor things. Will try this out later

Sov-trotter:
One thing that we need here is be able to specify whether we directly stream to twitch or to a local ip/port. What can be the possible syntax for that along with `livestream=true` flag?

Wikunia:
We might want to just have `livestreamto` and then options `nothing` as default, `twitch` and `local` or something.

Wikunia:
That is pretty neat @Sov-trotter 
As you mentioned it's currently a hassle to stop the stream.
I was wondering whether it's possible to not show the output of ffmpeg such that one can easily start a new stream but you probably had the same idea already.
Another thing is it would be great to start the stream before rendering is completed if that is possible. 

Some kind of a bug:
```
[h264 @ 0x1ba5fc0] Invalid NAL unit 0, skipping.=01:01:26.30 bitrate= 621.6kbits/s dup=0 drop=1579 speed=17.7x    
    Last message repeated 2 times
[h264 @ 0x1ba5fc0] Invalid NAL unit 0, skipping.=01:01:35.26 bitrate= 621.6kbits/s dup=0 drop=1583 speed=17.7x    
    Last message repeated 3 times
[h264 @ 0x1ba5fc0] Invalid NAL unit 0, skipping.=01:01:44.13
```
I get this in the console and sometimes the stream goes dark and after a couple of seconds it's back to normal.

Killing the ffmpeg process also doesn't kill the stream directly on obs which I find weird but I also don't know how it works exactly in the background whether it's really "live" or a bit cached.

src/Javis.jl
```diff
@@ -184,7 +186,10 @@ Renders all previously defined [`Object`](@ref) drawings to the user-defined `Vi
 - `tempdirectory::String`: The folder where each frame is stored
     Defaults to a temporary directory when not set
 - `liveview::Bool`: Causes a live image viewer to appear to assist with animation development
-- `liveview::Bool`: Livestream the rendered animation to the local network
+- `livestreamto::Union{Symbol, Nothing}`: Livestream the rendered animation to the local network
+- `address::String` : Address for the livestream if streaming to local
+- `port::Int` : Port number for the livestream
+- `twitch_key::String` : The twitch streaming key for your account
```
Wikunia, line 192: I was wondering whether we might want to have some kind of struct for it or a livestream function. Just to keep the number of parameters smaller but not sure yet.

Sov-trotter:
>I was wondering whether it's possible to not show the output of ffmpeg such that one can easily start a new stream but you probably had the same idea already.

Yeah that's fixed in the latest commit(had to let the loglevel flag).

>Another thing is it would be great to start the stream before rendering is completed if that is possible.

I do want to try that out, and would love to have that as a future milestone. Ref : https://trac.ffmpeg.org/wiki/StreamingGuide#Pointtopointstreaming `Saving a file and Streaming at the same time` section.

>Some kind of a bug;I get this in the console and sometimes the stream goes dark and after a couple of seconds it's back to normal.

It's related to the encoder which skips some frames then,I guess we can go around this by tuning some parameters?

>Killing the ffmpeg process also doesn't kill the stream directly on obs which I find weird but I also don't know how it works exactly in the background whether it's really "live" or a bit cached.

Yeah OBS has been unfriendly in that regard and has held this PR for a few days after which I finally decided to get things right at Javis's side first. We obviously expect OBS to update the preview automatically, but unfortunately there's a 1-2 minute delay before it does that. I discussed it with @TheCedarPrince yesterday. I did file an issue at obs https://github.com/obsproject/obs-studio/issues/4753

src/Javis.jl
```diff
@@ -169,6 +169,8 @@ end
         tempdirectory="",
         liveview=false,
         livestream=false,
```
Wikunia, line 171: should be the same as the description below

Wikunia (commented):
Just some minor comments again :wink: 
Will test it out later again!

src/Javis.jl
Sov-trotter: Yeah. That's needed for sure. We can have a constructor viz. `setup_livestream()` and pass the `struct` to the render function

Wikunia:
Is it possible to get the id of the ffmpeg process through julia and kill that one before starting a new stream?
Currently one can't run it twice, i.e. after making a change of the code.

Sov-trotter:
Yeah. That's the plan I currently have. The tricky thing is we need to know the PID of the ffmpeg process. `@schedule` doesn't do that though. We can do something like `pidof ffmpeg` and `kill <pid>`, but that will be confusing in the case where the user has multiple ffmpeg processes running.

Wikunia:
Yes maybe we can grep for the `pathname` as well to be sure that it's actually the one from Javis.

Sov-trotter:
`ps aux | grep ffmpeg | grep test.gif | awk '{print $2}'
`

Sov-trotter:
I think rather than pathname we can use some other parameter to check for the correct process since pathname can be different eg: `stream_loop` is a good parameter

Wikunia:
Yeah but as we know the pathname it should be easy, right?
Another thing that might exist is to add some info field in the command which does nothing which we can use to find the command.
Not sure whether `ffmpeg` supports things like that.
Also yes `stream_loop` is probably enough.

Sov-trotter:
>Yeah but as we know the pathname it should be easy, right?

Yeah. But I am also thinking from the perspective of another `cancel_stream()` function which will be independent of the `render` function, which shouldn't take in the pathname as it won't be userfriendly

src/javis_viewer.jl
```diff
@@ -287,17 +287,38 @@ function _pluto_viewer(video::Video, frames::Int, objects::Vector)
...
     port::Int
     twitch_key::String
 end
 
+"""
+    setup_stream(;livestreamto, address, port, twitch_key)
+
+Sets up the livestream configuration
+"""
 function setup_stream(;livestreamto::Symbol = nothing, address::String = "0.0.0.0", port::Int = 8080, twitch_key::String = "")
```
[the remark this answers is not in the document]
Wikunia: And I would prefer to write `setup_stream(:local)`.

src/javis_viewer.jl
```diff
@@ -286,3 +286,95 @@ function _pluto_viewer(video::Video, frames::Int, objects::Vector)
...
+    livestreamto::Symbol
+    address::String
+    port::Int
+    twitch_key::String
+end
+
+"""
+    setup_stream(;livestreamto, address, port, twitch_key)
+
+Sets up the livestream configuration
```
Wikunia, line 311: Should show the defaults.

src/javis_viewer.jl
```diff
@@ -286,3 +286,95 @@ function _pluto_viewer(video::Video, frames::Int, objects::Vector)
...
+    return "Livestream Cancelled!"
+end
+
+"""
+    _livestream(streamconfig, framerate, width, height, pathname)
+
+Internal method for livestreaming 
+"""
+function _livestream(streamconfig::StreamConfig, framerate::Int, width::Int, height::Int, pathname::String)
+    livestreamto = streamconfig.livestreamto
```
Wikunia, line 343: should cancel a stream if one is running, right? 
Otherwise the user has to always cancel it manually.

Wikunia (commented):
I know it's not ready for review yet but if you appreciate some feedback now I made some minor comments.

src/javis_viewer.jl
Sov-trotter: Yeah that was the original plan, but I realised, ffmpeg does that even more gracefully(internally) as compare to our `cancel_stream` method.

Wikunia: How? :wink:

Sov-trotter:
@Wikunia @TheCedarPrince  I think I have been able to solve the problem! The latest commit includes the fix. I would like if you guys can try it out and give feedback when we meet tomorrow?

Wikunia:
Thanks @Sov-trotter 
Feels like I still have to cancel the stream manually and get an infinite amount of error messages otherwise.

Sov-trotter:
Are the error messages similar to the ones you shared above? Apparently I don't get any.
Could you also share the example you're trying?

Wikunia:
```julia
using Javis

function ground(args...)
    sethue("black") # pen color
end

function object(p = O, color = "black")
    sethue(color)
    circle(p, 25, :fill)
    return p
end

function path!(points, pos, color)
    sethue(color)
    push!(points, pos) # add pos to points
    circle.(points, 2, :fill) # draws a circle for each point using broadcasting
end

function connector(p1, p2, color)
    sethue(color)
    line(p1, p2, :stroke)
end

myvideo = Video(1000, 1000)

path_of_red = Point[]
path_of_blue = Point[]

Background(1:70, ground)
red_ball = Object(1:70, (args...) -> object(O, "red"), Point(100, 0))
act!(red_ball, Action(anim_rotate_around(2π, O)))
blue_ball =
    Object(:same, (args...) -> object(O, "yellow"), Point(200, 80))
act!(blue_ball, Action(anim_rotate_around(2π, 0.0, red_ball)))
Object(1:70, (args...) -> connector(pos(red_ball), pos(blue_ball), "black"))
Object(1:70, (args...) -> path!(path_of_red, pos(red_ball), "red"))
Object(1:70, (args...) -> path!(path_of_blue, pos(blue_ball), "blue"))

streamconfig = setup_stream(:local)
render(myvideo; pathname = "tutorial_1.mp4", streamconfig)
```

if you run that two or sometimes three times (after changing a color or something but that is probably irrelevant) I get these errors:
```
[h264 @ 0x87ec00] Error splitting the input into NAL units.
Error while decoding stream #0:0: Invalid data found when processing input
[h264 @ 0x89b840] Invalid NAL unit size (1851780139 > 1442).
[h264 @ 0x89b840] Error splitting the input into NAL units.
Error while decoding stream #0:0: Invalid data found when processing input
[h264 @ 0x8b84c0] Invalid NAL unit size (-1644236136 > 626).
[h264 @ 0x8b84c0] Error splitting the input into NAL units.
Error while decoding stream #0:0: Invalid data found when processing input
[h264 @ 0x8d5140] Invalid NAL unit size (2016978250 > 394).
[h264 @ 0x8d5140] Error splitting the input into NAL units.
Error while decoding stream #0:0: Invalid data found when processing input
[h264 @ 0x8f1dc0] Invalid NAL unit size (-1424218937 > 407).
[h264 @ 0x8f1dc0] Error splitting the input into NAL units.
```
in a never ending stream. Everything works nicely when I run `cancel_stream()` in between each time though.

Wikunia:
That works!

Sov-trotter:
@Wikunia I am able to reproduce the error. 
Not cancelling the stream before creating the new one was creating a race condition.

Wikunia:
Oh I will not be able to stream two things at the same time via different ports, right?

Sov-trotter:
We can tweak the `cancel_stream()` method to support that by declaring a data structure that stores features like(address, filename, port etc.) and only kill the process if they match?

Wikunia:
Yeah maybe we keep this so for now to have a working version for it. I'm currently testing it for a video I want to make.
Might be a good way to have an instant updated preview :smile:

Sov-trotter:
image
How to set this up in OBS's side(for people trying this out)

src/javis_viewer.jl
```diff
@@ -286,3 +286,107 @@ function _pluto_viewer(video::Video, frames::Int, objects::Vector)
...
+    StreamConfig
+
+Holds the conguration for livestream, defaults to `nothing`
+
+#Fields
+- `livestreamto::Symbol` Livestream platform `:local` or `:twitch`  
+- `protocol::String` The streaming protocol to be used. Defaults to UDP
+- `address::String` The IP address for the `:local` stream(ignored in case of `:twitch`)
+- `port::Int` The port for the `:local` stream(ignored in case of `:twitch`)
+- `twitch_key::String` TWITCH API key for your account
```
TheCedarPrince, line 300: put this in a streaming struct file please.

src/javis_viewer.jl
```diff
@@ -286,3 +286,107 @@ function _pluto_viewer(video::Video, frames::Int, objects::Vector)
...
+Sets up the livestream configuration
+"""
+function setup_stream(livestreamto::Symbol = :local; protocol::String = "udp", address::String = "0.0.0.0", port::Int = 8080, twitch_key::String = "")
+    StreamConfig(livestreamto, protocol, address, port, twitch_key)
+end
+
+"""
+    cancel_stream()
+
+Kills the livestreaming process
```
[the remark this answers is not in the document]
TheCedarPrince: Add optional variable to kill online (i.e twitch processes) ffmpeg instances

TheCedarPrince (changes requested):
Some tweaks to adjust but otherwise looks REALLY great!

src/javis_viewer.jl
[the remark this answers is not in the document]
Sov-trotter: @Wikunia upon discussion with @TheCedarPrince I realise that there are many ways to improve this function.  Eg: We can use something like `lsof -i -P -n | grep ffmpeg`which scans all the listening ports on the local network rather than searching from all the active processes, but this won't work in case of twitch streaming, which can be solved by taking an optional stream config parameter into the function.

Also the current method of searching for ffmpeg streaming process creates problems with multiple streams(streaming to different animations/addresses/ports/protocols). Currently that isn't allowed as we cancel the ongoing stream before creating a new one to avoid race condition. If we want to support that in the future me might need to improve the searching methods for ffmpeg process.

Wikunia:
Have you tested streaming directly to twitch yet @TheCedarPrince ?

Wikunia:
@Sov-trotter what was the reason to change the default port and address? 
I get a `error writing trailer of udp://127.0.0.0:8081: Permission denied` now

Sov-trotter:
Yeah sorry I didn't mention this before committing. 
I am actually trying look for a default ip/port which has low probability of being used(we can put a check in the code if the port is taken but I still want the default to be the lesser used one)

Eg : I created jupyter notebook session and tried streaming to the same ip and port and it did work as long as the streaming protocol was not tcp. If it was tcp, it ended up crashing even the jupyter kernel and that's something I really want to avoid. 
Also `0.0.0.0:8080` is used quite frequently imo.

Wikunia:
How about 14015 for the closest approximation to "Javis" I can think of :smile: 
But I think using `0.0.0.0` makes sense as it's often used for this, right?

Sov-trotter:
I am not sure. But now I think leaving it to 0.0.0.0 is actually better than `127.0.0.1` since it's used quite often

Sov-trotter:
>av_interleaved_write_frame(): Permission denied
error writing trailer of udp://127.0.0.0:8081: Permission denied

Do you have something running at that IP/port?

Sov-trotter:
Also @TheCedarPrince @Wikunia it'd be nice if you can also review the workflow docs.

Wikunia:
> > av_interleaved_write_frame(): Permission denied
> > error writing trailer of udp://127.0.0.0:8081: Permission denied
> 
> Do you have something running at that IP/port because this works for me.

Nope I think it might be more the fact that it's `127.0.0.0` rather than `127.0.0.1`

docs/src/workflows.md
```diff
@@ -181,3 +181,42 @@ This will give you the ability to preview an animation with a slider:
 If you don't want to use the interactivity and simply display the output, set the `liveviewer` flag as `false` in the Javis `render` function and Pluto will display the gif:
 
 ![](assets/pluto_viewer.gif)
+
+
+## Livestreaming animations
+Javis brings in the support to pipe the rendered animations to streaming softwares(like OBS, VLC) or directly to platforms like Twitch or any alternatives that can read from a network stream.
```
Wikunia, line 187: should be `to streaming software` without the `s`

docs/src/workflows.md
```diff
@@ -181,3 +181,42 @@ This will give you the ability to preview an animation with a slider:
...
+
+- ### Streaming Locally
+Many eaducational streamers/teachers prefer using streaming tools like OBS/VLC to have more cretive control over thier streams. Javis livestreaming is tuned for that use case.
```
Wikunia, line 190: *eductational

Wikunia: *their

docs/src/workflows.md
```diff
@@ -181,3 +181,42 @@ This will give you the ability to preview an animation with a slider:
...
+The animation can be streamed to a local IP and port and can be viewed using any tool that can read from a network stream.
+
+Firstly, one needs to setup the `StreamConfig` object that holds configuration regarding the streaming process. Javis uses `udp://127.0.0.1:8081` as the default streaming protocol, address and port respectively. However you're free to use any options of your choice.
```
Wikunia, line 193: make sure to update this if you change the ip, port as this currently says 127.0.0.1

docs/src/workflows.md
```diff
@@ -181,3 +181,42 @@ This will give you the ability to preview an animation with a slider:
...
+On OBS:
+![](assets/local_obs.png)
+
+For tools like ffplay one can view the stream by typing `ffplay <protocol>://<address>:<port>`  in the terminal.
+
+eg: `ffplay udp://127.0.0.1:8081`
+
+- ### Streaming to Twitch
+Javis allows streaming animations directly to the [Twitch](twitch.tv) platform.
+All you need to do is pass the symbol `:twitch` and your twitch api key as arguments to the `stream_setup` method.
```
Wikunia, line 217: a link to how to obtain the key would be helpful

Wikunia (changes requested):
Some documentation changes.

TheCedarPrince:
@Sov-trotter - just let me know when you want me to take a look at this again and I can do a final pass and bounce it back to @Wikunia again for final signoff for merging. :)

Sov-trotter:
I have committed the changes requested by @Wikunia, so yeah you can share your feedback! :)

Wikunia:
Have you seen this error before @Sov-trotter : 
```
[gif @ 0x10cfd80] seek_frame_generic failed as this stream seems to contain no keyframes after the target timestamp, 1002 non keyframes found
```
It happened to me when I changed from 1000 frames to 1200.

Sov-trotter:
@Wikunia I can reproduce this. However the error takes place at the last frame(or after the last frame). 
REPRO:
- Remove the -loglevel error flag
-  set the number of frames to 1200

Also I didn't notice any flicker or frame drop when the error was thrown.  
Same for mp4 `[h264 @ 0x1b758c0] Invalid NAL unit 8, skipping.`

TheCedarPrince:
Should be ready for merging provided no formatting issues. Feel free to review @Wikunia , merge, and either you or me can tag a release. Great work @Sov-trotter :raised_hands: :tada: :tada:

Wikunia:
The changelog section seems to be missing

src/javis_viewer.jl
```diff
@@ -286,3 +279,124 @@ function _pluto_viewer(video::Video, frames::Int, objects::Vector)
...
+    end
+
+    command = [
+        "-stream_loop", # loop the stream -1 times i.e. indefinitely
+        "-1",
+        "-r", # frames per second
+        "$framerate",
+        "-an",  # Tells FFMPEG not to expect any audio
+        "-loglevel", # show only ffmpeg errors
+        "error",
```
Wikunia, line 362: should this depend on the `ffmpeg_loglevel` of the `render` function?

Wikunia (commented):
Just one minor comment :wink:

src/javis_viewer.jl
Sov-trotter: I had actually thought of that. But since streaming is synchronous, having it is set to `error` by default is beneficial incase the stream fails

Wikunia: Oh yeah didn't realize the other one is `panic`. We might wanna have some `loglevel` struct later with different possibilities at some point. But it's okay for now.

PR #340, opened 2021-06-01
test on 1.6

### LINKED ISSUE #339: Test with julia 1.6
Currently CI tests only for julia 1.4 and 1.5. I think it'd be more reasonable to include julia 1.6(and maybe remove julia 1.4)

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #339 

Testing on 1.6 and 1.4 as 1.4 is the oldest version we support at this moment.


Wikunia:
Seems like there is an issue with VideoIO as the duration on one of our test cases is off.

Wikunia:
Does this fail locally on your machine @TheCedarPrince ? Works for me and I would really like to test Javis 1.6. Seems like it uses the same VideoIO version and I have locally so unsure where the error might come from.

Wikunia:
Already done with #368

PR #348, opened 2021-06-15
fixing some bugs in the workflow docs

Wikunia:
Some documentation fixes in the workflow section:

- Removed `:balloon:` which didn't render
- Made a section bold instead of screaming
- Comment out the Twitch section instead of using an HTML comment which doesn't work


TheCedarPrince (approved):
LGTM

Issue #313, opened 2021-02-18
Predefined objects and improved user experience  


Wikunia:
**Is your feature request related to a problem? Please explain.**
I feel like writing anonymous functions is sometimes a pain and it doesn't contain much structure if everything is an `Object`.

**Describe the solution you'd like**
I'm wondering if it makes sense to have some types for `Line, Circle, Rectangle, Polygon` etc. and one can write:
```julia
myline = Object(1:100, Line(p1, p2))
act!(myline, Action(1:10, appear(:draw)))
```
such that the line gets drawn (see #218 )
and maybe also 
```julia
myline = Object(1:100, [Line(p1, p2), Line(p2, p3)])
act!(myline, Action(1:10, appear(:draw)))
```
such that we can concatenate line segments easily.

Additionally each object could have some properties like, `start_point`, `end_point` for line and `radius` and `center` for Circle.

Then one would be able to write:
```julia
myline = Object(1:100, Line(p1, p2))
act!(myline, Action(1:10, appear(:draw)))
mycircle = Object(1:100, Circle(myline.start_point, 20))
```
where the `start_point` field is an anonymous function which has access to the start point of the line. 

The main point for `Line` and the other objects is also that we can define them ourselves on top of Luxor and be able to specify all possible options to it like `color`, `width` for which one currently needs to write anonymous functions.

Old functionality should not be destroyed by this but just extended and it should be simple for the user to define their own objects.

**Describe alternatives you've considered**
I considered using 
```julia
Line(1:100, p1, p2)
```
but it feels better to have the `Object` around it which takes care of the frames.

This might also somehow make it easier to work with #130 .

@TheCedarPrince What do you think and feel free to rename this issue as naming is hard :smile:

Sov-trotter:
I think this task in general has tons of possibilities. I am not yet ready with a concrete design yet. For now I think we need a general data structure/API to store not just the start point but any information in general, hence a mutable `metadata` field attached with each object.

Wikunia:
Oh yeah for sure. This was just an example. Each `drawingType` should have its own fields and possibilities to store such information

codejaeger:
Hey @Wikunia , @TheCedarPrince I was thinking about how this could be done and came up with the following

* Having a `metadata::Any[]` field seems important to me. It might seem useless for a `Line` or `Circle` where having a `Dict` to store the positions or radius arguments might suffice but take the example where one of the  argument itself is a `Dict`.
[...]

PR #350, opened 2021-06-23
Predefined objects and new actions

### LINKED ISSUE #218: Drawing a shape line by line
It's currently not really supported to draw a shape line by line. This can go from the drawing a rectangle line by line to the Hilbert curve.

It makes sense to have another `appear` function like `appear(:line)` which uses the polyportion functionality to make this effect.

**Additional context**
As an example
hilbert

Sov-trotter:
## PR Checklist

**Link to relevant issue(s)**
Closes #218 #313 

This implements the appear{:draw} method and predefines the `(args...) -> ...` declaration for some basic shapes. 
The medata implementation here is temporary as I store everything in a `Dict ` in the `object.result` field.

### DIFF
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -57,7 +57,6 @@ include("structs/Object.jl")
 include("structs/Transitions.jl")
 include("structs/Action.jl")
 
-
 """
     Line
 
@@ -97,10 +96,12 @@ include("luxor_overrides.jl")
 include("backgrounds.jl")
 include("svg2luxor.jl")
 include("morphs.jl")
+include("structs/Creations.jl")
 include("action_animations.jl")
 include("javis_viewer.jl")
 include("latex.jl")
 include("object_values.jl")
+include("shorthands.jl")
 
 """
     projection(p::Point, l::Line)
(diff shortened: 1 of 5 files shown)


Sov-trotter:
ease_pos_test
I have been thinking of having `:draw` action for a bunch of basic shapes. Obviously the methods for each differ quite alot. 
What other basic shapes can we include @TheCedarPrince @Wikunia both for `:draw` and object declarations?

Wikunia:
The idea was to be able to use this for all polygons such that the user can write something like the space filling curve in #218 by simply using line functions inside a user defined function.

Sov-trotter:
There are a few things here:
- The current way of drawing lines and circles is quite hacky. For a line I have just drawn it from point to point(using the get_interpolation) method. 
- For a circle, I have created a clipping mask using a pi and increased it's angle according to the interpolation factor. 
- Now we can go on creating drawing actions for different shapes. But what we really want to do it beable to do that for any polygon for that matter. https://juliagraphics.github.io/Luxor.jl/stable/polygons/#Luxor.polyportion is what we need to look at and bridge it with the interpolation factor. To get increasing portions of any polygon iteratively

Issue #141, opened 2020-09-01
Drawing functions of Luxor should return a point or bounding box


Wikunia:
**Is your feature request related to a problem? Please explain.**
Currently `circle` doesn't return a point such that one has to build an own circle function when one wants to animate another object based on the position of the circle center.

**Describe the solution you'd like**
It would be awesome to just have overridden methods of `circle`, `rect` and probably others which return either a specific point like center for circle and top left for rect. 

**Describe alternatives you've considered**
Currently one has to write a own method but it might be convenient to have it built into Javis.

TheCedarPrince:
I think this is a theme that is related to #130 in the sense of making easy short-hands.

Wikunia:
Best is probably to return the origin by default as in general it makes sense to draw at the origin and then translate it later. 
So if no return is done just `return O` by default.

Issue #130, opened 2020-08-31
Convenience Object Position Functions


TheCedarPrince:
**Is your feature request related to a problem? Please explain.**
Currently, `Javis.jl` requires one to _absolutely_ position an object with pixel locations. This is not user friendly. Furthermore, to make objects drawn on a canvas move in relation to one another is currently not possible. The only way to make objects move in concert would require the user to carefully note each object manually. How can we make this easier?

**Describe the solution you'd like**
An idea might be to use short hand expressions to define general functions similar to what `manim` does. An example of `manim` syntax is as follows:

```python
from manimlib.imports import *

class ReplacementTransformExhibit(Scene):
    def construct(self):
        square = Square().shift(LEFT)
        circle = Circle().shift(RIGHT)
        self.play(ReplacementTransform(square, circle))
```

Which renders the following animation:

![manim_anim](https://user-images.githubusercontent.com/29561456/91678138-ba8d9100-eb12-11ea-9b31-96a9773132b9.gif)

> Link: https://eulertour.com/docs

Of note in this example are the short hands, `LEFT` and `RIGHT`. This shifts the `Square` object from the `LEFT` and the `Circle` to the `RIGHT`. The `ReplacementTransform` shifts these two shapes properly while transforming between the two objects (i.e. square to circle).

Based on this example, it is apparent to see that the circle and square have a default placement in this `Scene`. I think clever usage of both defaults and supported polygons are important.

Perhaps we could make something like a `shapes.jl` file which contains short hand definitions for common objects like circles, squares, diamonds, triangles, etc. based on `Luxor.jl` such that one could define a short hand shape like `Circle == circle(O)`. Furthermore, perhaps we could use `morphs.jl` to create short hands like `TransformTo(circle, square)` to transform a circle to a square at a default speed.

PR #356, opened 2021-07-03
[GSoC'21] Convenience methods for objects

### LINKED ISSUE #313: Predefined objects and improved user experience

Sov-trotter:
## PR Checklist

**Link to relevant issue(s)**
Closes #313 #130 #141 

For now I have implemented abstractions for basics objects like, 
- [x] Circle
- [x] Line
- [x] Box
- [x] rect
- [x] triangle/polygon
- [x] ellipse
- [x] star
....

So these objects basically wrap the `(args...)->func` declarations. It would be useful to have a method for a vector of such shapes. 
This also has some potential in the direction of having internal flow of metadata.

For avoiding conflicts with luxor's and javis' already existing basic shapes, the methods have been named `J<SHAPE>` eg: `JCircle`,`JStar`
- [ ] With convenience methods, support minimizing expressivity in frames:


Sov-trotter:
Test Script: https://gist.github.com/Sov-trotter/2df6181fd2ca2d2f121ecd7be490c2aa

Sov-trotter:
@TheCedarPrince  @Wikunia  I can use a review at this point :)

Sov-trotter:
@TheCedarPrince  I remember us talking about what if someone wants to define some custom anonymous functions that are much detailed in functionality.  Eg: the `path!` function here https://github.com/Wikunia/Javis.jl/blob/master/docs/src/tutorials/tutorial_1.md#mapping-an-orbit-

For that I have created a macro viz. `JShape` which allows one to wrap such instructions in a block.
```
somepath = Object(1:70,  Javis.@JShape begin
                                    sethue("blue")
                                    circle.(<out points>, 20, :fill)
                                end 
)

```

src/shorthands/JBox.jl
```diff
@@ -0,0 +1,53 @@
...
+    box(pt, width, height, cornerradius, action=action)
+end
+
+"""
+    JBox(cornerpoint1::Point, cornerpoint2::Point; color="black", action=:stroke, vertices=false)
+
+Create a box (rectangle) between two points and do an action.
+Use vertices=true to return an array of the four corner points: bottom left, top left, top right, bottom right.
+"""
+JBox(cornerpoint1::Point, cornerpoint2::Point; color="black", action=:stroke, vertices=false) = (args...) -> _JBox(cornerpoint1, cornerpoint2, color, action, vertices)
```
Wikunia, line 24: This currently limits the functionality as one can't use the `change` Action, as that needs keyword arguments, right?
I think:
```julia
JBox(cornerpoint1::Point, cornerpoint2::Point; color="black", action=:stroke, vertices=false) = (args...; kcolor=color, kaction=action, kvertices=vertices) -> _JBox(cornerpoint1, cornerpoint2, kcolor, kaction, kvertices)
```
should work? 
Maybe even have `cornerpoint` in there to have full flexibility.

src/shorthands/JBox.jl
```diff
@@ -0,0 +1,53 @@
...
+
+"""
+    JBox(points::Array; color="black", action=:stroke, vertices=false)
+
+Create a box/rectangle using the first two points of an array of Points to defined opposite corners.
+Use vertices=true to return an array of the four corner points: bottom left, top left, top right, bottom right.
```
Wikunia, line 30: Regarding: #141 we might want to return the upper left corner by default?

src/Javis.jl
```diff
@@ -424,6 +424,28 @@ for func in names(Luxor; imported = true)
...
+include("shorthands/JBox.jl")
+
+# Ellipse
+include("shorthands/JEllipse.jl")
+
+# Star
+include("shorthands/JStar.jl")
+
+# Poly
+include("shorthands/JPoly.jl")
```
Wikunia, line 447: I would remove the comments here besides the `#shorthand declarations` one as they don't provide any value.

Wikunia (commented):
Thanks for creating this PR @Sov-trotter I think it will simplify a lot of things in the future. Especially the `JShape` is a great addition. Is it possible to maybe make that accept keyword arguments as well? For example 
```julia 
@JShape begin
   ....
end color=:black radius=20
```

Sov-trotter:
> Is it possible to maybe make that accept keyword arguments as well?

Yeah. That is possible. But I am not sure how we can use those arguments inside since we just evaluate the expression inside the anonymous function. https://github.com/Wikunia/Javis.jl/pull/356/files#diff-357a561d1b0f8639d39256acccb59064e0cdde58dc13775c9d8f9e3bcefbbf01R18

Wikunia:
Oh yeah I meant `JShape`. Yeah I'm not that knowledgeable in macros :confused: 
Wondering whether one can parse the arguments and than evaluate those before calling `eval`
Something like:

```julia
macro JShape(body::Expr, expr)
    kwargs = (; expr.args[1] = expr.args[2])
    expr = quote
        (args...) -> begin 
            esc(
                for (key, value) in enumerate($kwargs)
                   @eval (($key) = ($value))
                end
            )
            eval($body)
        end
    end
    esc(expr)
end
```

which obviously doesn't work :smile: 

In your case one would need to parse the vars and values out of the `begin end` block and replace them with the kwargs, right?

Sov-trotter:
>Wondering whether one can parse the arguments and than evaluate those before calling eval

I think this should work. 
```julia
@JShape begin
   sethue(color)
   circle(radius)
   line(O)
end color=:black radius=20
```
I that what you have in mind?

Sov-trotter:
>In your case one would need to parse the vars and values out of the begin end block and replace them with the kwargs, right?

No. The arguments go inside the begin end block and are parsed along with the expression. So they are not essentially arguments but expressions.

src/shorthands/JBox.jl
Sov-trotter: So this needs to be done for all the shapes right?

Wikunia:
> ```julia
> @JShape begin
>    sethue(color)
>    circle(radius)
>    line(O)
> end color=:black radius=20
> ```
> 
> Is that what you have in mind?

yes that's what I thought.

Wikunia:
> > In your case one would need to parse the vars and values out of the begin end block and replace them with the kwargs, right?
> 
> No. The arguments go inside the begin end block and are parsed along with the expression. So they are not essentially arguments but expressions.

But one would need to extract them to write the `kwargs` and would need to change the right hand side to actually use the kwarg value instead of the default.

src/shorthands/JBox.jl
Wikunia: Yes :wink:

Sov-trotter: Okay So this works!
Is there any specific reason for having "k" pre-fixed to the argument names.
` kcolor=color, kaction=action, kvertices=vertices) `

I tried with `color=color, action=action, vertices=vertices) ` and it works.

Wikunia: I was wondering which one is taken then but as it's a different scope it's probably fine :smile:

Sov-trotter:
>>@JShape begin
   sethue(color)
   circle(radius)
   line(O)
end color=:black radius=20
Is that what you have in mind?
yes that's what I thought.

src/shorthands/JBox.jl
Sov-trotter: Do we need to compute that? Because there's no direct way of  doing that?

Wikunia:
I like your way of having it inside but assumed it's harder. One can parse them as keyword arguments though, so that shouldn't be the biggest problem

Sov-trotter:
@Wikunia I think kwargs can now be passed. The trick was to not consider body as a seaparate argument but as a keyword argument. 

The catch here is that on need to pass all the required arguments before the begin...end block. So something like:
```julia
@JShape color=:black radius=20 begin
     sethue(color)
     circle(radius)
     line(O)
  end 
```
Since eval() takes in a list of expressions. So if the misc. arguments are passed after the block, they will not be available for use/might throw an error for the first frame where object.func is called

docs/src/tutorials.md
```diff
@@ -14,10 +14,11 @@ Their level of difficulty starts from a beginner `Javis` user to advanced usages
 Currently, these tutorials are available:
 
 - [**Tutorial 1: Making Your First `Javis` Animation!**](tutorials/tutorial_1.md) - a step by step guide to making your first animation with `Javis`.
-- [**Tutorial 2: What Are Actions?**](tutorials/tutorial_2.md) - an introduction to the most important part of `Javis` - the `Action` object.
-- [**Tutorial 3: Rendering `LaTeX` with `Javis`!**](tutorials/tutorial_3.md) - a simple tutorial on how to render `LaTeX` in `Javis` animations.
-- [**Tutorial 4: Do You Know Our Mascot? Learn about Transitions and Actions!**](tutorials/tutorial_4.md) - a fun tutorial to make our mascot and learn about the `Action` and `Transition` type methods.
-- [**Tutorial 5: Taming the Elements**](tutorials/tutorial_5.md) - how to use `change` to grow or shrink arbitrary objects and using `Javis` with other Julia packages.
-- [**Tutorial 6: Using Animations.jl to Create something with more Pep!**](tutorials/tutorial_6.md) - an advanced tutorial to make your animations more interesting.
+- [**Tutorial 2: Using Shorthands in Javis!**](tutorials/tutorial_6.md) - Using shothands in `Javis`.
```
Wikunia, line 17: link seems wrong

src/shorthands/JBox.jl
```diff
@@ -0,0 +1,126 @@
...
+    sethue(color)
+    box(pt, width, height, cornerradius, action)
+    return Point(pt.x - width / 2, pt.y + height / 2)
+end
+
+"""
+    JBox(cornerpoint1::Point, cornerpoint2::Point; color="black", action=:stroke, vertices=false)
+
+Create a box (rectangle) between two points and do an action.
+Use vertices=true to return an array of the four corner points: bottom left, top left, top right, bottom right.
```
Wikunia, line 46: This seems to be off. It only returns `top left` now

src/shorthands/JPoly.jl
```diff
@@ -0,0 +1,46 @@
...
+    end
+)
+```
+"""
+macro JShape(body...)
+    expr = quote
+        (args...) -> $_JShape($body)
+    end
+    esc(expr)
+end
```
Wikunia, line 46: Do we want to have this in an extra file?

test/shorthands.jl
```diff
@@ -0,0 +1,130 @@
...
+@testset "Shorthands" begin
+    function ground(args...)
+        background("white")
+        sethue("black")
+    end
+
+    Background(1:70, ground)
+
+    line = Object(Javis.JLine(Point(100, -250)))
+    line1 = Object(Javis.JLine(Point(-350, 350), Point(100, -250)))
```
Wikunia, line 12: We want to export those shorthands

Wikunia (commented):
Just some minor things for now. I'll have a look at `JShape`

Wikunia:
@Sov-trotter do you think this will work?
```julia
macro JShape(body::Expr, expr...)
    leftside =  Expr(:tuple, Expr(:parameters, [Expr(:kw, e.args[1], e.args[2]) for e in expr]...), Expr(:(...), :args)) 
    out = Expr(:(->), leftside, body)
    esc(out)
end
```
It's from here: https://discourse.julialang.org/t/macros-to-create-anonymous-function/65426
Currently it seems to be working just fine with the `change` function and to provide not only constants but given values as well like:
```julia
color = "blue"
object = Object(1:100, Javis.@JShape begin
    sethue(color)
    circle(O, radius, :fill)
 end color=color radius=20 )
 act!(object, Action(1:100, change(:radius, 20=>40)))
```

Sov-trotter:
Ah. This works neatly!
Tahnks for making the discourse post!

src/shorthands/JBox.jl
Sov-trotter: I think that was in the issue? https://github.com/Wikunia/Javis.jl/pull/356#discussion_r663476254

Wikunia: Yeah just the docstring needs to be updated

Sov-trotter: Oh sorry.

Sov-trotter: BTW Is this necessary? Since box() will return all the points of the box as a  list which seems much better.

Wikunia: It's useful for using the `pos` function.

src/shorthands/JLine.jl
```diff
@@ -0,0 +1,22 @@
...
+    JLine(p1::Point, p2::Point, color="black")
+
+Make a line between two points, pt1 and pt2 and do an action.
+Returns the final point of the line
+"""
+JLine(p1::Point, p2::Point; color = "black") =
+    (args...; color = color, p1 = p1, p2 = p2) -> _JLine(p1, p2, color)
+
+"""
+    JLine(pt::Point)
```
Wikunia, line 17: the color is missing in this docstring. I would also like to have `linewidth` for `setline` default should be 1.

Wikunia (commented):
Experimenting a bit with this and ODEs a bit. Have just one small request for JLine for now. I can add that myself though as well.
ode
@TheCedarPrince that is the start of what I was telling you about

src/shorthands/JBox.jl
```diff
@@ -0,0 +1,126 @@
...
+        cornerpoint2 = cornerpoint2,
+        color = color,
+        action = action,
+        vertices = vertices,
+    ) -> _JBox(cornerpoint1, cornerpoint2, color, action, vertices)
+
+"""
+    JBox(points::Array; color="black", action=:stroke, vertices=false)
+
+Create a box/rectangle using the first two points of an array of Points to defined opposite corners.
```
TheCedarPrince, line 67: Could you add a documentation reference to `Points` here?

src/shorthands/JShape.jl
```diff
@@ -0,0 +1,21 @@
...
+    JShape(body...)
+
+Creates a custom shape based on the luxor instructions in the begin...end block
+```julia
+somepath1 = Object(@JShape begin
+                                    sethue(color)
+                                    poly(points, action, close= true)
+                                end action = :stroke color ="red" radius = 8 
+)
+```
```
TheCedarPrince, line 11: Could you run JuliaFormatter on the code in this example? Somehow seems a bit off to me.

test/shorthands.jl
```diff
@@ -0,0 +1,126 @@
+video = Video(800, 800)
+
+@testset "Shorthands" begin
```
TheCedarPrince, line 3: Either in this test set or in the Layer test set, could you test running J-Objects within a Layer?

TheCedarPrince (commented):
Looks really great! I edited the Tutorial some to make it a bit less technical but kept the same outline and general content. Also, the only major comment I left is I would just like to see a few tests for J-Objects being used with a Layer. Otherwise, when done, I think this is ready for merging!

src/shorthands/JBox.jl
Wikunia: You mean `Point` ? But that is in Luxor so referencing it is unfortunately not that simple.

TheCedarPrince: Ah right - never mind! Resolving this.

test/shorthands.jl
Wikunia: done in the Layer tests with JShape.

Wikunia:
Would be lovely if you could check my docstring changes and overall structural change of some methods @TheCedarPrince

Sov-trotter:
https://github.com/Wikunia/Javis.jl/pull/356/commits/973e95c73f2af40e5f2543d72b0e5d7d23c40e7a#diff-d35b42f4d138de274461c0f7ed00bfc89a3e90294bc64e30feae10a1f3ef2e84R13-R46 this is a really interesting way of using convenience functions.

Wikunia:
> [973e95c#diff-d35b42f4d138de274461c0f7ed00bfc89a3e90294bc64e30feae10a1f3ef2e84R13-R46](https://github.com/Wikunia/Javis.jl/commit/973e95c73f2af40e5f2543d72b0e5d7d23c40e7a#diff-d35b42f4d138de274461c0f7ed00bfc89a3e90294bc64e30feae10a1f3ef2e84R13-R46) this is a really interesting way of using convenience functions.

Haha yeah I really like that way to be honest. For this one it's a bit basic but can get quite handy sometimes.

TheCedarPrince (approved):
Looks good to me! One thing I am curious about is if we can do something down the road similar to `matplotlib` to handle the kwargs found in different functions - especially the J-Objects here (see **kwargs here: https://matplotlib.org/devdocs/api/_as_gen/matplotlib.pyplot.plot.html). I wonder if Documenter is smart enough to render documentation for a function's kwargs. I'll investigate. 

Also, I was going to say that I wasn't exactly a fan of the 1., 2., etc. docstrings at first, but I think they make a lot of sense now. Especially looking at `JEllipse` - it makes a lot of sense. I think this is set to merge - great work @Sov-trotter and @Wikunia !

P.S. Made a small commit to fix the docstring of the `@JShape` macro.

Wikunia:
Thanks for the review. Can you clarify a bit what you mean by the keyword argument thing?

PR #366, opened 2021-07-19
Fix link to Luxor tutorial

cormullion:
I recently re-organized the documentation of Luxor (along `divio` lines) so any direct links you've made to individual sections of the Luxor docs, like the one in this PR, will probably 404.

Naturally I'm very sorry I've caused you extra work. :) But on the plus side, I probably won't reorganize them again for a while...


Wikunia:
Thanks @cormullion we normally want to link to the stable version though, right?

cormullion:
Good point!

Wikunia:
Thanks again!

PR #368, opened 2021-07-21
videoIO v0.9 & test on Julia v1.6

Wikunia:
## PR Checklist

If you are contributing to `Javis.jl`, please make sure you are able to check off each item on this list:

- [x] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [x] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [x] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [x] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [x] Did I properly add test dependencies to the `test` directory (if applicable)?
- [x] Did I check relevant tutorials that may be affected by changes in this PR?
- [x] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**
Closes #325 
Closes #328 

**How did you address these issues with this PR? What methods did you use?**
I've worked along the lines of https://github.com/JuliaIO/VideoIO.jl/blob/master/CHANGELOG.md to make the change from v08 to v0.9 of VideoIO. Additionally I added to test on v1.6 of Julia as this wasn't possible before (#340)

PR #370, opened 2021-07-21
docstring fix or anim_translate + change when from and to are given

### LINKED ISSUE #367: [BUG] anim_translate feels broken. How to fix?
**Describe the bug**

As an example by @simeonschaub I had a look at `anim_translate` again and have the feeling that it's kinda broken but I don't have a simple fix in mind. 

The code:
```julia
using Javis
function ground(args...)
    sethue("black") # pen color
end
function object(p=O, color="black")
    sethue(color)
    circle(p, 25, :fill)
    return p
end
myvideo = Video(500, 500)
Background(1:70, ground)
upper_ball = Object(1:70, (args...)->object(O, "red"), Point(0, -100*√3))
left_ball = Object(1:70, (args...)->object(O, "green"), Point(-100, 0))
right_ball = Object(1:70, (args...)->object(O, "blue"), Point(100, 0))
act!(left_ball,  Action(anim_translate(Point(2 * 100, 0))))
act!(right_ball, Action(anim_translate(Point(2 * -100, 0))))
render(myvideo; pathname="triangle.gif")
```

It currently produces:

javis_VNNDI7e

but expected would have been:

triangle

especially based on our docstring:
```julia
- `anim_translate(x::Real, y::Real)` define end point of translation
- `anim_translate(tp::Union{Object,Point})` define end point with [`Object`](@ref) or `Point`
```

Now the first question is: Do we want to define the end point or not?
- It seems cool to have this possibility and @simeonschaub clearly expected that behavior (not sure whether this was based on the docstring or intuition)

The problem here is the following:

```julia
left_ball = Object(1:70, (args...)->object(O, "green"), Point(-100, 0))

act!(left_ball,  Action(1:70, anim_translate(Point( 100, 0))))
act!(left_ball,  Action(1:70, anim_translate(Point( 0, 100))))

render(myvideo; pathname="counter_example.gif")
```

The intuition from before seems to break down when two `anim_translate` functions are acted on at the same time.
What should this do? If we have the end point idea from before it should somehow be at two places at the same time...
Should this just be avoided? Or it it better to see `anim_translate` as forces and the end point is `Point(0, 100)` ?
In that case the user has to keep track of the current position. 

Another option is to have a function `anim_translate_to` which uses the position at the start of the action as a reference point and updates the position at the end of the action. 

Looking forward to your inputs @TheCedarPrince @Sov-trotter @cormullion

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #367 

`anim_translate` was broken and this is one possible fix. 
It was broken in the sense that we mentioned in the docstring that we move to a specific position but actually we move by a vector/into a direction. 
Additionally I've updated the behavior when `from` and `to` are both given such that it assumes we are already at `from` because at that stage it is unknown where we are... Previously it assumed we are at the origin but this should be clearer now.


TheCedarPrince (commented):
Things look mostly good to me here @Wikunia ! The only issue I had is I am a little confused by what you meant by the docstring:

```
- `anim_translate(tp::Union{Object,Point})` define direction and length of the translation vector with [`Object`](@ref) or `Point`
```

What are the two ways you are suggesting here with calling this function? I figure one was `anim_translate(Point(x, y))` but what was the other way of calling this function?

TheCedarPrince:
Also ran the test suite and am still running into the VideoIO error for mp4 on my end here:

```
Dancing circles (mp4): Test Failed at /home/src/Projects/javis/test/animations.jl:88
  Expression: isapprox(VideoIO.get_duration("dancing.mp4"), 30.0, atol = 0.1)
   Evaluated: isapprox(28.0, 30.0; atol = 0.1)
Stacktrace:
 [1] macro expansion
   @ /home/src/Projects/javis/test/animations.jl:88 [inlined]
 [2] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl
:1151 [inlined]
 [3] top-level scope
   @ /home/src/Projects/javis/test/animations.jl:64
```

TheCedarPrince:
If you can clarify that part of `anim_translate` as I mentioned earlier, feel free to merge. Thanks!

Wikunia:
> Also ran the test suite and am still running into the VideoIO error for mp4 on my end here:
> 
> ```
> Dancing circles (mp4): Test Failed at /home/src/Projects/javis/test/animations.jl:88
>   Expression: isapprox(VideoIO.get_duration("dancing.mp4"), 30.0, atol = 0.1)
>    Evaluated: isapprox(28.0, 30.0; atol = 0.1)
> ```

That is expected as it is another PR #368

PR #373, opened 2021-07-24
[GSoC'21] Morphing should change the object

### LINKED ISSUE #286: Morphing should change the object
Currently it's a bit hard to morph an object and then change the color of the morphed object afterwards. This is due to the fact that morphing doesn't really change the object. 

I think we should consider actually changing the object function by the `morph_to` function. The problem here is that it would make it impossible to work with `liveview`. Therefore, we might want to have it inside the `ObjectSetting` instead such that we could overwrite the function without loosing the user defined function.

Another possibility would be that we can somehow return an object from `act!` but that feels weird as well.

Sov-trotter:
## PR Checklist

**Link to relevant issue(s)**
Closes #286 

So the way this currently works is, by replacing the object's function with the function to which the object is supposed to be morphed.

So something like:
```julia
# inside the morph_to function
if frame == last(get_frames(action))
        object.func = to_func
    else
        # the morphing computation
        newpath()
        ....
        return morph_between(video, action, frame, from_polys, to_polys; do_action = do_action)
    end

```

[TEST SCRIPT](https://gist.github.com/Sov-trotter/6141ed9bed79bdb4a7b1a40b00227175)


test/viewer.jl
```diff
@@ -95,7 +95,16 @@ end
     frames = Javis.preprocess_frames!(objects)
 
     @test v.filename === "foo.png"
-    img = Javis._pluto_viewer(vid, length(frames), objects)
+    img = Javis._pluto_viewer(vid, length(frames)-1, objects)
```
Wikunia, line 98: does this mean that the pluto viewer will generally have problems? Like switching from the end frame of morphing back?
Maybe we should save the original function in the `Object` struct as well then and set it if it's not the last frame?

Wikunia (commented):
Do you mind running `format` over this PR and make a note in the changelog?

test/viewer.jl
Sov-trotter: Yeah. I breaks jupyter and javis viewer. Pluto will work since we iterate over a list of frames.

test/viewer.jl
```diff
@@ -95,7 +95,16 @@ end
     frames = Javis.preprocess_frames!(objects)
 
     @test v.filename === "foo.png"
-    img = Javis._pluto_viewer(vid, length(frames), objects)
+    img = Javis._pluto_viewer(vid, length(frames) - 1, objects)
```
Wikunia, line 98: The old version should work now, right?

Wikunia:
If it now works in Jupyter and the liveviewer I'm very happy with the PR. Then a single line in the `changelog` file would be nice I'll merge.

test/viewer.jl
Sov-trotter: Oh sorry. Forgot about that.

test/viewer.jl
```diff
@@ -96,6 +96,9 @@ end
 
     @test v.filename === "foo.png"
     img = Javis._pluto_viewer(vid, length(frames), objects)
+    objects = vid.objects
+    frames = Javis.preprocess_frames!(objects)
```
Wikunia, line 100: is this needed due to this PR?

Sov-trotter: No. This is to test if the pluto viewer generates frames correctly.

Wikunia: I was wondering why it wasn't there before

Sov-trotter: I removed the black lines. https://github.com/Wikunia/Javis.jl/blob/master/test/viewer.jl#L94-L95

Wikunia:
Great! Two things:
- `Changelog.md` needs the update
- If you would add a test case where this functionality is actually tested that would be awesome

Sov-trotter:
So https://github.com/Wikunia/Javis.jl/pull/373/commits/6b5f4cf037853056b8efc803103de209c8588e37#diff-29df3566a6e16bde78716afd075fb04532e77f70ffad38d58cad1661b797f124R62 was creating a bug.

Now when we try to morph the object further, it restores the original function hence, if we morph a rect to a circ, then a circ to a star, it restores the rect when morphing for star has to begin since all of these actions act on the same object.

testmorph

Sov-trotter:
The latest commit fixes this 
testmorph

src/morphs.jl
```diff
@@ -42,6 +42,7 @@ function morph_to(to_func::Function; do_action = :stroke)
         _morph_to(video, object, action, frame, to_func; do_action = do_action)
 end
 
+const PREVIOUS_MORPH_ACTION = Array{Action,1}()
```
Wikunia, line 45: I assume this isn't working when you more several objects at the same time.
I think this should be part of `opts` so per object

test/morphing.jl
```diff
@@ -156,3 +156,63 @@ end
...
+    back = Background(1:200, ground)
+    star_obj = Object(1:200, abox)
+    act!(star_obj, Action(10:20, morph_to(acirc)))
+    act!(star_obj, Action(30:40, anim_translate(Point(100, -100))))
+
+    # this is also a bug with the morph_to function
+    # the star is formed at the incorrect position
+    # the the origin is shifted and neve restored back
+    # you can see this in the result of anim_translate actions
+    # fix this at some point!
```
Wikunia, line 180: Maybe make this into a `TODO:` such that we find this at a later stage.

Wikunia:
Is this ready from your side @Sov-trotter ?

Sov-trotter:
Yeah!

PR #374, opened 2021-07-24
Different morphing modes

Sov-trotter:
## PR Checklist

This PR introduces some manim like modes for morphing 
https://azarzadavila-manim.readthedocs.io/en/latest/animation.html#transform

### DIFF
--- a/src/Shape.jl
+++ b/src/Shape.jl
@@ -221,7 +221,7 @@ function print_basic(s::Shape)
 end
 
 
-function prepare_to_interpolate(from_shape, to_shape)
+function prepare_to_interpolate(from_shape, to_shape, style)
     # match number of points for outer polygon
     from_outer, to_outer = match_num_points(from_shape.points, to_shape.points)
 
@@ -236,8 +236,12 @@ function prepare_to_interpolate(from_shape, to_shape)
     end
 
     # rotate outer polygon
-    rotate_i, _ = compute_shortest_morphing_dist(from_outer, to_outer)
-    new_from_outer = circshift(from_outer, -rotate_i + 1)
+    if style == :short
+        rotate_i, _ = compute_shortest_morphing_dist(from_outer, to_outer)
+        new_from_outer = circshift(from_outer, -rotate_i + 1)
+    elseif style == :long
+        new_from_outer = circshift(reverse!(from_outer), floor(-length(from_outer)/4))
+    end
 
     new_from_holes = Vector{Vector{Point}}()
 
--- a/src/morphs.jl
+++ b/src/morphs.jl
@@ -37,9 +37,9 @@ circle_obj = Object(11:20, acirc)
 act!(circle_obj, Action(:same, morph_to(astar)))
 ```
 """
-function morph_to(to_func::Function; do_action = :stroke)
+function morph_to(to_func::Function; style = :short, do_action = :stroke)
     return (video, object, action, frame) ->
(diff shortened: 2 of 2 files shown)


src/Shape.jl
```diff
@@ -236,8 +236,12 @@ function prepare_to_interpolate(from_shape, to_shape)
     end
 
     # rotate outer polygon
-    rotate_i, _ = compute_shortest_morphing_dist(from_outer, to_outer)
-    new_from_outer = circshift(from_outer, -rotate_i + 1)
+    if style == :short
+        rotate_i, _ = compute_shortest_morphing_dist(from_outer, to_outer)
+        new_from_outer = circshift(from_outer, -rotate_i + 1)
+    elseif style == :long
```
Wikunia, line 242: doesn't this depend on the ordering of the initial polygons? Would it make sense to circshift according to shortest first and then reverse that?

Wikunia (commented):
Just a minor comment here. Let me know when I should do a proper full review

Sov-trotter:
@Wikunia since the goal here is more visual appeal, should we provide some default easing function(apart from linear()) in this PR??

Wikunia:
I think it makes sense to have the easing separated from the function to have it more composable. The default for all actions should be `linear()` I think. We might want to create predefined actions similar to predefined objects in the future where we can tackle this. What do you think?

Sov-trotter:
>I think it makes sense to have the easing separated from the function to have it more composable. 

Understood. I was just of the opinion, that in longer running videos it would rather be easy for the user if the small morphs had atleast some visual appeal by default. 

>We might want to create predefined actions similar to predefined objects in the future where we can tackle this.

Sure yeah. I had thought of that while tackling predefined objects. There are a few things to look into before that. 
First is making the Javis Recipes(which are obviously actions) more "available". I mean to say that the users don't have to create overloads like: 
`(video, object, action, rel_frame) -> _translate(video, object, action, rel_frame)`
and then go on to define their custom actions in `_foo` function. 

Maybe a macro is handy here. 
So this itself accounts for predefined actions(for recipes/internally)

Secondly we can have some predefined actions but at present I still don't know we to wrap around? Would like to hear what ideas you have.

Wikunia:
Let's discuss that tmr in our meeting :wink:

Sov-trotter:
Self TODO : Check if morphing modes work with #286

src/Shape.jl
```diff
@@ -221,7 +221,7 @@ function print_basic(s::Shape)
 end
 
 
-function prepare_to_interpolate(from_shape, to_shape)
+function prepare_to_interpolate(from_shape, to_shape, style)
```
Wikunia, line 224: do you mind adding a small docstring to this function?

Wikunia (commented):
One minor thing

Wikunia:
Still not a fan of `:long` how about `:shortest_dist` and `:visually_appealing` ? 
Maybe also add how it works in the docstring of `morph_to`.

Wikunia:
We've decided to call those modes `mode1, mode2, ...` and have animations in the docs for the different morphing modes.
Additionally it would be good to have arguments to parameterize some of the modes like the one implemented in this PR

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Javis.jl - Changelog
 
+## Unreleased
+- Add a manim-like :long morhping mode
```
Wikunia, line 4: Should be updated here as well

PR #385, opened 2021-08-06
fixed layer dimensions when not given explicitly

Wikunia:
## PR Checklist

The `@JLayer` macro was not working when a video wasn't evaluated before the macro as the `CURRENT_VIDEO` needed to exist to make it work when the user didn't specify the dimensions of `@JLayer`.
Now it is decided at runtime which fixes that issue. 

Additionally I've created a `preprocess_frames!` function which takes in the `video` instead of objects and calls it with objects and layers. I've also restructured the testcase for layers as with #338 the `CURRENT_VIDEO` will be reset after a render and one shouldn't specify a `Video` outside of a `testset`.

Do you mind reviewing this pr @Sov-trotter ? 
Additionally a comment to this line `@test_throws ErrorException Javis.get_layer_frame(vid, l5, 6)` in the tests would be great as I don't understand why it should/does fail :smile:


test/layers.jl
```diff
@@ -175,7 +171,7 @@ video = Video(600, 600)
     rm("layer_test.gif")
 
 
-    Javis.CURRENT_VIDEO[1] = vid
```
Sov-trotter, line 178: I think this was to reset the CURRENT_VIDEO but that wasn't a clean way. This is better now.

test/layers.jl
```diff
@@ -214,6 +210,6 @@ video = Video(600, 600)
 
     act!(l5, Action(opacity_anim, setopacity()))
 
+    Javis.preprocess_frames!([vid.objects..., Javis.flatten(vid.layers)...])
     @test_throws ErrorException Javis.get_layer_frame(vid, l5, 6)
```
Sov-trotter, line 214: So this method should throw an error due to `act!(l5, Action(opacity_anim, setopacity()))`
   `Frame range for the layer's action might be missing`

Sov-trotter:
I think rest of the changes are good! Happy to have them merged.

PR #387, opened 2021-08-07
Livestreaming v2

### LINKED ISSUE #345: Make livestreaming live
With #337 we'll have a way to stream a rendered video to twitch or locally. Unfortunately it isn't really live, as the video needs to be rendered beforehand. 

I'm wondering whether it's feasible to stream single images instead to have an actual live view which can be like 5fps but at least it's live. That way we could figure out the current position of the live stream and could update some changes live without restarting the animation. 
In the future I would love to stream events like presidential elections live on a map. It would take new data from a source every couple of minutes and then starts an animation with it.

**Additional context**
This can be similar to something like this then: https://www.twitch.tv/dutchsinseofficial

Sov-trotter:
## PR Checklist

Closes #345 
As discussed this makes livestreaming more live, this time by streaming the frames rather than a gif. 
Also allows to stream specific frames by passing in a `UnitRange`


Sov-trotter:
@TheCedarPrince @Wikunia I think this is ready for review

Sov-trotter:
@TheCedarPrince @Wikunia I think this is ready for review

src/Javis.jl
```diff
@@ -275,6 +275,11 @@ function render(
         tempdirectory = mktempdir()
     end
 
+    if streamconfig != nothing
```
Wikunia, line 278: use `===` or `!==` when comparing if a value is nothing

src/javis_viewer.jl
```diff
@@ -342,20 +343,37 @@ Internal method for livestreaming
...
 
     if livestreamto == :twitch && isempty(twitch_key)
         return error("Please enter your twitch stream key")
     end
 
+    stream_frames = config_frames == :all ? frames : config_frames
+    stream_filecounter = 1
+    for frame in stream_frames
+        frame_image = convert.(RGB, get_javis_frame(video, objects, frame; layers = layers))
+        if !isempty(tempdirectory)
```
Wikunia, line 368: what happens if tempdirectory is empty? It just generates the frames but can't use it?

Wikunia (changes requested):
To be honest I expected something quite different with #345 
Currently this does still precomputes all image frames before starting to stream. 
Live streaming should mean that it starts streaming as soon as the first frame got computed.

Wikunia:
I'm unsure what broke when merging it with master. Do you have an idea @Sov-trotter ? Feel free to revert the changes I made.

Sov-trotter:
>Live streaming should mean that it starts streaming as soon as the first frame got computed.

Yeah. I understand. Just taking small steps here. Will get to that too. :)

src/javis_viewer.jl
Sov-trotter: No tempdir is created before if it's empty (as before)
https://github.com/Wikunia/Javis.jl/pull/387/files/56156371a5657544c632a3cb6faa2b39681e1fe7#diff-763ffddd0fe1ffb533564007035f5947812905676655f155ce1ee4357eb53c5eR275

Sov-trotter:
>Live streaming should mean that it starts streaming as soon as the first frame got computed.

I think we cna talk about how to do this in a cleaner fashion. Because subsequent frames are generated in a serial fashion in a for loop.

I wonder how to have ffmpeg read that. We can obviously overwrite one image again and again but that might cause a race condition. 

https://newbedev.com/can-you-stream-images-to-ffmpeg-to-construct-a-video-instead-of-saving-them-to-disk might be of help

src/javis_viewer.jl
Wikunia: Yeah what I mean is: it should never be empty so we don't need the if. The problem is if we have an mp4 as the pathname without a temp directory but want to stream it will fail, right?

Sov-trotter:
Hey @Wikunia  @TheCedarPrince after some more experimentation here's what I found:
- I think it's more reasonable to have the the the livestreaming "semi-live" as currently in this PR or lazy streaming? Not sure if I am using the term "lazy" correctly?

- As I mentioned, there are two ways of making it actually live first is using `image2pipe` and other is streaming a single image and overwriting it in julia. 
- image2pipe works. One work around I found was since it doesn't support `-stream_loop` we can create our own loop in julia that iterates to large values that creates a new ffmpeg process everytime. 
- This works but is a bit laggy on ffplay, while obs works smoothly.
- However I find this to be slow since get_javis_frames is called for every iteration of the video. Hence using image2pipe generally is not a good idea since that will require us to feed in frames for every iteration hence computationally expensive.
- The above point also applies to overwriting the frame image, since we expected some caching of the frames to take place but that doesn't quite happen without using `stream_loop`. 
- Also creating a new image and replacing the previous frame with the new one isn't computationally great either. 

What we can do is remove the images from temp directory when `cancel_stream` is called.

PR #390, opened 2021-08-11
allow color interpolation in change

### LINKED ISSUE #363: Animate change of color
I couldn't really figure out a way to fill color in a shape gradually(not instantly).
eg: Something like 
```julia
act!(star_obj, Action(41:50, change(:do_action, :fill)))
act!(star_obj, Action(41:50, change(:color, "red")))
``` 
Fills color in the 41st frame rather than animating the filing.

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Partially closes #363

I've changed the standard interpolation in the `change` function to use Animations.jl now.
That means we interpolate a single time to get the `t` value using Animations.jl which is based on the easing function but then also use Animations.jl to interpolate using the `t` value the actual value that we want to change. 

Additionally I saw that `JBox` only allows strings as colors which shouldn't be the case so I fixed that as well.


TheCedarPrince (approved):
LGTM - I like the syntax that we now have with this. Feel free to merge.

I wonder if we should add this to a tutorial somewhere. I could see redoing Tutorial 7 with J-Objects and this syntax in mind. But that is another issue. :) 

Merge it!

TheCedarPrince (changes requested):
Oh actually, can you go ahead and use this PR to bump the release version? That way we can have this released and close another issue or two.

TheCedarPrince (approved):

PR #399, opened 2021-08-14
`JImage` Implementation for J-Objects

### LINKED ISSUE #388: Add Functionality to Easily Add Images to J-Objects
After some fun experiments today, I realized that we do not have a way to easily add images to J-Objects! I was inspired by @Wikunia's blog post on [The Chasing Problem](https://opensourc.es/blog/javis.jl-examples-series-the-chase-problem/) to "fix" it. :wink: In short, I was trying to add a clipped image to the JCircle function and realized it was rather unclear how to do this outside of base Javis. Although I could, I think that this could be useful to add into J-Objects and much easier. Here is what I made:

chase

Here is the syntax I created for a function call:

`JCircle(dog_positions[i], 27; img = readpng("tiny_doggo.png"))`

Then, here is what I changed in base Javis to the `JCircle` function:

```julia
JCircle(
    center::Point,
    radius::Real;
    color = "black",
    linewidth = 1,
    action = :stroke,
    img = nothing,
) =
    (
        args...;
        center = center,
        radius = radius,
        color = color,
        linewidth = linewidth,
        action = action,
        img = img,
    ) -> _JCircle(center, radius, color, linewidth, action, img)
```

and here is the internal method:

```julia
function _JCircle(center, radius, color, linewidth, action, img)
    sethue(color)
    setline(linewidth)

    if !isnothing(img)
        circle(center, radius, :stroke)
        circle(center, radius, :clip)
        placeimage(img, center, centered = true)
    else
        circle(center, radius, action)
    end

    return center
end
```

I considered writing an external function outside of Javis to do this, but it felt like too much work for such a small new feature addition. Then, I thought about creating a new base Javis function. That didn't make much sense as it is not much different from the base `JCircle` J-Object implementation. So, I added it to the base Javis `JCircle` definition.

What do you think @Wikunia and @Sov-trotter ? I tested out the above changes and none of the tests we had in place failed using this new functionality.

TheCedarPrince:
## PR Checklist

**Link to relevant issue(s)**
Closes #388 

This is a proposed implementation idea for adding images to J-Objects easily. I was inspired by some of the conversation in #388 and wanted to create something easily like

```julia
function JImage(....)
     placeimage(....)
end
```

But could not get this syntax to work as when you have two separate objects - say JCircle which clips a circle shape and JImage - the clipping does occur but as JCircle does not "see" JImage, the expected output (i.e. the JImage is not clipped in a circle shape) does not occur. This PR introduces syntax such that you can write:

```julia
Object(
    JImage(
        pos = O,
        img = readpng("my_picture.png"),
        centering = true;
    ),
)
```
Which acts as a fully qualified Javis object (i.e. you can perform actions with it). Furthermore, clipping and outlining can be supported with the following kwarg syntax:

```julia
Object(
    JImage(
        pos = O,
        img = readpng("my_picture.png"),
        centering = true;
	shapeargs = (pt = O, r = 40, action = :clip),
	shape = circle,
    ),
)
```

So this picture will be cliipped by passing shapeargs and a Luxor shape function.

jimage_test

_Picture of cat from Twitch Artist ChargedAsylum and owner AftonSteps_

## Disclaimer

I have not added tests or docs as of yet as I am waiting to see what the thought is with this possible syntax.

Pinging @Wikunia and @Sov-trotter for thoughts!

### DIFF
--- a/src/shorthands/JImage.jl
+++ b/src/shorthands/JImage.jl
@@ -1,0 +1,17 @@
+function _JImage(pos, img, centering, clipargs, clipshape)
+    if clipshape != false
+        clipshape(clipargs...)
+    end
+    placeimage(img, pos, centered = centering)
+    return pos
+end
+
+JImage(pos::Point, img, centering = true; clipargs = (), clipshape = false) =
+    (
+        args...;
+        pos = pos,
+        img = img,
+        centering = centering,
+        clipargs = clipargs,
+        clipshape = clipshape,
+    ) -> _JImage(pos, img, centering, clipargs, clipshape)
--- a/src/Javis.jl
+++ b/src/Javis.jl
@@ -682,6 +682,7 @@ include("shorthands/JEllipse.jl")
 include("shorthands/JStar.jl")
 include("shorthands/JPoly.jl")
 include("shorthands/JShape.jl")
+include("shorthands/JImage.jl")
 
 export render, latex
 export Video, Object, Background, Action, RFrames, GFrames
@@ -696,7 +697,7 @@ export act!
 export anim_translate, anim_rotate, anim_rotate_around, anim_scale
 
 export @Frames, prev_start, prev_end, startof, endof
-export JBox, JCircle, JEllipse, JLine, JPoly, JRect, JStar, @JShape
+export JBox, JCircle, JEllipse, JImage, JLine, JPoly, JRect, JStar, @JShape
 
 # custom override of luxor extensions
 export setline, setopacity, fontsize, get_fontsize, scale, text


src/shorthands/JImage.jl
```diff
@@ -1,17 +1,17 @@
-function _JImage(pos, img, centering, clipargs, clipshape)
-    if clipshape != false
-        clipshape(clipargs...)
+function _JImage(pos, img, centering, shapeargs, shape)
+    if shape != false
+        shape(shapeargs...)
```
Sov-trotter, line 3: One minor thing:
I think we should mention(in the docstring) that the shape should be a valid Luxor shape.
eg: rect, box, poly etc.

cormullion: "valid Luxor shape" I'm not aware of this concept. :) Do you mean a built-in Luxor function - eg one that constructs paths, or one that generates points for polygons...? (Not trying to be awkward, just I get worried in case I'm not providing the tools you folks need for your cool constructions!)

Sov-trotter: >Do you mean a built-in Luxor function - eg one that constructs paths
Yeah! We need shape to be a valid luxor function, but a user may interpret it as entering a full shape name eg: rectangle instead of `rect`.

Wikunia: We mean any kind of function either their own or more likely a Luxor drawing function. Mostly those which provide the clipping action (so yeah a drawing function 😂)

cormullion: OK, yes. Perhaps any function that has an `action` parameter/keyword...

Sov-trotter: Another thing I want to mention is that `shape` should by typed to `Method`. 
eg: If there's a variable `circle = Object(...)`, and user passes circle as an argument, it will throw the `Objects of type Javis.Object are not callable` error and maybe use a try catch block to give out a better error message?

src/shorthands/JImage.jl
```diff
@@ -0,0 +1,17 @@
...
+    end
+    placeimage(img, pos, centered = centering)
```
Wikunia, line 5: would be nice to be able to scale the image before it gets placed

TheCedarPrince: Agreed! Added it in

TheCedarPrince:
@Wikunia and @Ved-Mahajan - I think the interface for `@JImage` is mostly stabilized to something that looks like this:

```julia
Object(
    JImage(
        pos = O,
        img = readpng("my_picture.png"),
        centering = true;
	shapeargs = (pt = O, r = 40, action = :clip),
	shape = circle,
        scaleargs = .5
    ),
)
```

What is left to do is add documentation and tests as of this moment. Anything else you guys can think of?

src/shorthands/JImage.jl
```diff
@@ -0,0 +1,19 @@
+function _JImage(pos, img, centering, shapeargs, shape, scaleargs)
+    if shape != false
+        shape(shapeargs...)
+    end
+    scale(scaleargs)
+    placeimage(img, pos, centered = centering)
+    return pos
+end
+
+JImage(pos::Point, img, centering = true; shapeargs = (), shape = false, scaleargs = 1) =
```
Wikunia, line 10: I think `shape = nothing` would be a better default.

Wikunia: The function needs a docstring which should probably explain clipping as part of shapeargs.

src/shorthands/JImage.jl
```diff
@@ -0,0 +1,19 @@
+function _JImage(pos, img, centering, shapeargs, shape, scaleargs)
+    if shape != false
+        shape(shapeargs...)
+    end
+    scale(scaleargs)
```
Wikunia, line 5: I would really like to have automatic scaling by checking the size of the shape and that of the image. 
For the shape it would be awesome to be able to call it with `:clip` / `:path` automatically but it seems `action` isn't always a keyword argument which makes that a bit hard. Any idea on that @cormullion ?

Wikunia (changes requested):
Some small comments for improvement and clarification/docstrings.
Can do another review after test cases are added.

Wikunia:
Sorry I kinda missed your example from the PR text. I think that looks great. I was wondering whether we could combine it with the other `J` objects such that one has a dispatch option to pass those as the shape. That might make it easier for the user and we would be able to call the appropriate `:clip` by default and determine the size by calling it with `:path` I assume as in those `J` Objects the action is a keyword argument. I would still keep the current form you have though to keep the flexibility. As a user I would find the other one easier to write though.
Something like:

```
Object(
    JImage(
        pos = O,
        img = readpng("my_picture.png"),
        centering = true;
	shape = JCircle(0, 40),
    ),
)
```

src/shorthands/JImage.jl
cormullion: yes, following on from my question (https://github.com/Wikunia/Javis.jl/pull/399#discussion_r689130301) not all graphic construction functions are similar in having an action keyword. We'd have to compile a list and see which ones could be updated...

Wikunia: Are you interested in standardizing the functions though such that all have an option which provide using the `action` as a kwarg?

cormullion: Perhaps we can make a list of possible functions as an issue at Luxor.jl. It might be easy or not, not too sure how many are affected...

Wikunia: https://github.com/JuliaGraphics/Luxor.jl/issues/174

src/shorthands/JImage.jl
TheCedarPrince: That makes sense. Done.

src/shorthands/JImage.jl
TheCedarPrince: since this spawned off another issue, shall we resolve this comment @Wikunia and @cormullion ?

src/shorthands/JImage.jl
```diff
@@ -1,13 +1,21 @@
...
     placeimage(img, pos, centered = centering)
     return pos
 end
 
-JImage(pos::Point, img, centering = true; shapeargs = (), shape = false, scaleargs = 1) =
+"""
+    JImage(pos::Point, img, centering = true; shapeargs = (), shape = nothing, scaleargs = 1)
+
+Place a given image at a given location as a `Javis` object.
+Images can be cropped to different shapes and scaled to different sizes while being placed.
```
Wikunia, line 14: make sure to explain the arguments in more detail i.e. based on the docstring I don't really know what `img` is. Is it a filename or an image matrix?

Wikunia:
You should be able to use Luxor 2.16 now and use `action` as a kwarg. Then scaling could be done automatically.

src/shorthands/JImage.jl
TheCedarPrince: I decided to close this conversation as I do not think automatic scaling is feasible as all shapes can define a bounding box differently within Luxor. 
To find that bounding box for a shape depends on the arguments provided by such a shape.

src/shorthands/JImage.jl
TheCedarPrince: Fixed

TheCedarPrince:
This is nearly ready for merging, but somehow the CI is failing and I am having trouble running the test suite locally on my own machine.
Otherwise, docs have been added, tests have been made, and the functionality is working.
Should also explain what `scaleargs` can expect from `scale`.

Thoughts @Wikunia ?

src/shorthands/JImage.jl
cormullion: You can find a BoundingBox for a lot of "shapes" using eg `pathtopoly()`:

```julia
@draw begin
    circle(O, 100, :stroke)
    circle(O, 100, :path)
    bbox = BoundingBox(pathtopoly()[1])
    box(bbox, :stroke)
end
```

If you can convert a shape to a polygon, then you can get a bounding box that way. Are there any other shapes in Luxor that you'd like BoundingBox to work on?

test/shorthands.jl
```diff
@@ -147,3 +147,95 @@ video = Video(800, 800)
...
+
+    circle_img = Object(
+        1:5,
+        JImage(
+            O,
+            readpng("refs/dispatch.png"),
+            true;
+            shape = circle,
+            shapeargs = nothing,
+            scaleargs = 1,
```
Wikunia, line 168: Maybe add another test case with different shapeargs and different scale args.

src/shorthands/JImage.jl
```diff
@@ -0,0 +1,66 @@
...
+
+# Arguments
+- `pos::Point`: Where to place the image inside a shape
+- `img::String`: Expects the path to an image
+- `centering::Bool`: Centers the object at `pos`
+- `shapeargs`: Arguments to be passed to a given shape type
+- `shape`: A Luxor shape function such as `circle`, `box`, etc.
```
Wikunia, line 50: would have `shape` before `shapeargs` and maybe add an example which shows what `shapeargs` and `scaleargs` do?

Wikunia (commented):
Just some minor things. I assume #455 fixes your test cases

TheCedarPrince:
Hey @Wikunia , yes, that would help here as well as @cormullion pointing out a nice method that can help with automatic scaling!
Planning to hopefully look at this a bit today!

cormullion:
I think using bounding boxes might be an easy way to fit images into shapes. Here's some pseudo code:

```julia

# determine  BoundingBox:

bbox = BoundingBox(box(O, 100, 100)   # a simple rect

center = Point(60, -20)
radius = 120
bbox = BoundingBox(box(center, 2radius, 2radius)) # or a circle

bbox = BoundingBox(star(O, 100, 6, 0.5, vertices=true)) # or a poly

bbox = BoundingBox(first(pathtopoly())) # or the current path

# get image

img = readpng(...)

# calculate a scale factor, after determining whether you're fitting the 
# image (to see all of it) or filling (to occupy all the shape):

imagefit = true
if imagefit
    boxside = max(boxwidth(bbox), boxheight(bbox))
    imageside = max(img.width, img.height)
else
    boxside = min(boxwidth(bbox), boxheight(bbox))
    imageside = min(img.width, img.height)
end 
scalefactor = boxside/imageside

# draw clipping shape here

# finally place the image 
@layer begin
    translate(boxmiddlecenter(bbox))
    scale(scalefactor)
    placeimage(img, centered=true)
end
```

Not sure if this reliably works in all situations or in your Javis framework though... :)

TheCedarPrince:
Hey @cormullion , 
Just responding more directly here to your thoughts.
That pseudocode looks really great and actually very much like what I was hoping was available.
I just wasn't sure how to get the bounding box of "shape" functions.
As of right now, I need to double check but basically in my mind, _any_ Luxor "shape" function (like, `star`, `box`, `circle`, `ellipse`, `poly`, etc.) should be able to support getting a bounding box. 
Then, a generalized scale factor could be calculated to either fill or fit the picture (or a user could provide their own scaling should they so choose).

TheCedarPrince:
Hey @cormullion , 

So, I ran into the problem that I have been running into with different shape type functions.
That is, for example, `box`, will always return a vector of Points that can be used for constructing a bounding box.
However, a `circle`, will only return a `Bool` statement which cannot be used to build a `BoundingBox` to my knowledge.
Is there any way to construct a `BoundingBox` for shapes such as these?
Thanks!

cormullion:
>"I don't want a Boolean!"  😃😃😃😃

Unfortunately `circle()` talks to Cairo which internally generates a set of four Bezier curves but returns nothing, so there aren't any coordinates to return... 

So I've added a return statement to `circle()` that consists of a tuple of top left and bottom right corners of the bounding box of a  circle:

```julia
julia> circle(O, 10, :path)
(Point(-10.0, -10.0), Point(10.0, 10.0))
```

So you can now go:

```julia
@draw begin
    pt1, pt2 = circle(O, 100, :stroke)
    box(BoundingBox(pt1, pt2), action=:stroke)
end 
```

or even

```julia
@draw begin
    circle(O, 100, :stroke)
    bbox = BoundingBox(circle(O, 100))
    box(bbox, :stroke)
end
```

and similarly for `ellipse` and `rect`:

```julialang
@draw begin
    bbox = BoundingBox(ellipse(O, rand(1:200), rand(1:200), :stroke))
    box(bbox, :stroke)
end
```

I suppose these are breaking changes (if anyone is relying on circle returning a Boolean...!). Perhaps we'll go for Luxor 3.0!

Please check it out on the current Luxor!

TheCedarPrince:
@Wikunia - I have a potentially radical new proposal based on some thoughts you had earlier for this PR:

What if instead of adding a `JImage` function, I instead create a dispatch around the existing class of J-Shapes (i.e. `JBox`, JCircle`, etc.) to enable automatic scaling of images and image placement?
What do you think?
I personally think this would make the syntax a lot less clunky and code a lot easier to read.

TheCedarPrince:
- [X] JBox
- [ ] JCircle
- [ ] JEllipse
- [X] JPoly
- [ ] JRect
- [X] JStar

Hey everyone!
After much revision and review, I did decide to try overhauling the syntax I was proposing for `JImage`!
Instead, I decided what would be best is to build on the syntax of the J-Objects.
Therefore, we can get rather understandable syntax that looks something like this:

```julia
Object(
    31:45,
    JPoly(
        [Point(-100, 0), Point(0, -100), Point(100, 0), Point(80, 100), Point(-80, 100)],
        "white",
        1,
        :fill,
        true,
        false,
        "../test/refs/dispatch.png",
        :inset,
    ),
)
```

As I was working through this syntax, it made me wonder if @cormullion , would you be opposed to me making a Luxor PR to allow image placement inside of shapes such as `Circle`, `Rect`, and more?
It's a bit odd as this syntax would be specific to only Javis so it would feel nice to merge this functionality upstream and we pull it in from Luxor.
What do you think?
Feel free to look at the file, "src/shorthands/JBox.jl` for the wrapping mechanism I use.

@Wikunia , what do you think of this syntax?
It would also work for one to provide their own scaling factor if the options of `:clip` or `:inset` are not satisfactory.

Finally, here is code and a gif demonstrating some of the functionality:

```julia
Object(1:15, JBox(O, 200, 200, "white", :clip, false, "../test/refs/dispatch.png", :inset))

Object(
    16:30,
    JStar(O, 200, "white", 1, 5, 0.5, 0, :fill, false, "../test/refs/dispatch.png", :clip),
)

Object(
    31:45,
    JPoly(
        [Point(-100, 0), Point(0, -100), Point(100, 0), Point(80, 100), Point(-80, 100)],
        "white",
        1,
        :fill,
        true,
        false,
        "../test/refs/dispatch.png",
        :inset,
    ),
)
```

test

Also of note, this functionality will require the latest version of Luxor which I know will break morphing sadly..... :cry:

Wikunia:
Why does the latest Luxor break morphing?
The syntax looks good to me.  One problem I see with the positional arguments is that I have no idea what the two boolean arguments are. 
Seems like you haven't pushed your changes yet so I'm not sure whether they are introduced in this PR or whether they already exist in the `J` functions in general. 
Maybe we can change the image filename argument in such a way though that it is a keyword argument.

cormullion:
>  @cormullion , would you be opposed to me making a Luxor PR to allow image placement inside of shapes such as Circle, Rect, and more?

Haha, yes, probably (opposed), initially! 🤣🤣 Well, you'd have to persuade me - it would certainly be something that would need a fair bit of justification, to take account of the increased complexity, maintenance burden, introduction of (more) inconsistencies, testing, documentation, and so on., assuming there's no performance impact. But if you make a testable PR (with some benchmarks), we could discuss it...

TheCedarPrince:
Hey @Wikunia !

This is ready for review!
Let me know what you think of the syntax!
Temporarily, I have just been relying on positional arguments, but we can change that in the future after your review to make the functionality much more readable.

Here is sample code to use to test out this feature's functionality:

```julia
using Javis

video = Video(400, 400)
origin(Point(200, 200))
function ground(args...)
    background("black")
    sethue("white")
end

Background(1:90, ground)

Object(1:15, JBox(O, 200, 200, "white", :clip, false, "../test/refs/dispatch.png", :inset))

Object(
    16:30,
    JStar(O, 200, "white", 1, 5, 0.5, 0, :fill, false, "../test/refs/dispatch.png", :clip),
)

Object(
    31:45,
    JPoly(
        [Point(-100, 0), Point(0, -100), Point(100, 0), Point(80, 100), Point(-80, 100)],
        "white",
        1,
        :fill,
        true,
        false,
        "../test/refs/dispatch.png",
        :inset,
    ),
)

Object(46:60, JEllipse(O, 200, 100, "white", 1, :fill, "../test/refs/dispatch.png", :inset))

Object(61:75, JCircle(O, 100, "white", 1, :fill, "../test/refs/dispatch.png", :inset))

Object(76:90, JRect(O, 200, 100, "white", 1, :fill, "../test/refs/dispatch.png", :inset))

render(video; framerate = 5, pathname = "test.gif")
```

Also, @cormullion - that makes complete sense.
I will finish prototyping this functionality back here in Javis and then after we have something stable, I'll see about maybe making an example upstream PR to Luxor.
Thanks!

Wikunia:
I currently can't run your test as I get the error which also seems to appear in the test cases:
```
MethodError: no method matching Luxor.BoundingBox(::Bool)
```

src/shorthands/JCircle.jl
```diff
@@ -39,3 +39,58 @@ JCircle(p1::Point, p2::Point; kwargs...) =
     JCircle(midpoint(p1, p2), distance(p1, p2) / 2; kwargs...)
 
 JCircle(radius::Real; kwargs...) = JCircle(O, radius; kwargs...)
+
+"""
+TODO: Add Documentation!
```
Wikunia, line 44: currently I have no idea what `:inset` and `:clip` mean to be honest :smile: 
Seems like `:clip` positions the image somewhere else but I don't really have the option to move the image or have I?
Also for me a `scale_factor` is a number and not a symbol. Maybe `scale_type` or something but I would first need to know what they do to give my comment on that :smile:

Wikunia:
Make sure to update the `Project.toml` and specify that the master version of Luxor is needed or which version exactly.

PR #412, opened 2021-09-11
Scaling: linear

### LINKED ISSUE #355: Automatic converstion of Cartesian coordinates to Luxor coordinates.
This issue is for having a function which would transform the coordinates in Cartesian coordinate system to the Luxor coordinate system and will output `Points` which can then be animated by using `Animation` and `Javis`.

A function like `Cartesian2Points` which would take in a array or a tuple of Cartesian coordinates and would return a `Point` which can be readily used in `Javis`.
For example:
```julia
# Define the bodies
body1 = MassBody(SVector(0.0, 1.0, 0.0), SVector(5.775e-6, 0.0, 0.0), 2.0)
body2 = MassBody(SVector(0.0, -1.0, 0.0), SVector(-5.775e-6, 0.0, 0.0), 2.0)

G = 6.673e-11

system = GravitationalSystem([body1, body2], G)
tspan = (0.0, 1111150.0)

# run the simulation
simulation = NBodySimulation(system, tspan)
sim_result = run_simulation(simulation, saveat = 10)
sim_data = sim_result.solution.u

# extract the data
planet1_pos = [(sim_data[x][1, 1], sim_data[x][2, 1]) for x in 1:length(sim_data)]
planet2_pos = [(sim_data[x][1, 2], sim_data[x][2, 2]) for x in 1:length(sim_data)]

#Convert to Points
planet1_point = Point.((100 .* getindex.(planet1_pos, 1)), ((100 .* getindex.(planet1_pos, 2) .- 50)))
planet2_point = Point.((100 .* getindex.(planet2_pos, 1)), ((100 .* getindex.(planet2_pos, 2) .+ 50)))
```
The numbers `100` and `50` in the last two lines, if they are found automatically by taking into account the video size as fed into `Video` and the max and min coordinates in the Cartesian system.

Manually choosing the numbers to scale the coordinates to make the `Points`, which might not be efficient for a more complicated system.

**Additional context**

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #355 

One can now create a scale similar to scales in d3.js. (Currently only a simple linear scale).
The function `scale_linear` has two methods one for everything (mostly numbers) and one specific for points.


Wikunia:
You can now write something like:
```
pts = [Point(i,i) for i in 0:5]
mapping = scale_linear(O, Point(5, 5), Point(50, -50), Point(200, -200))
@scale_layer mapping begin
         circle.(pts, 0.1, :fill)
end
```
which maps the points to `Point(50, -50), Point(80, -80) ... Point(200, -200)`
and they'll have a radius of `3`. One has to be careful that the rectangles are congruent though as otherwise the circles would not be circles anymore :wink:

Wikunia:
@Ved-Mahajan You might be interested in this. Any thoughts?

Ved-Mahajan:
> @Ved-Mahajan You might be interested in this. Any thoughts?

Thank for the ping. If I understand correctly this is a transformation: `x --> m*x  + c` right?
In that case, that is exactly what I had in mind to have.
And I think the mapped points can be directly used in `Animation` to move the`Object`, which is good to have !!

Wikunia:
Yeah it's basically an affine transformation. One can't use translate by XY as from the outside it's still just a standard Luxor coordinate system but one probably wants use `change` for those points anyway.

Ved-Mahajan:
hmm yeah got it
I think one change needed would be invert the y axis.

One more thing needed would be to have that shift that @TheCedarPrince  did in that two body animation which was to adjust for the initial condition which is just addition by a constant which can be taken care by `c`.

Wikunia:
The flip in the y axis can be done with the above code and is done in the example 😉 
I'm not sure what you refer to with the two body animation. Shifting and scaling is what can be done with this linear scale functionality.

Ved-Mahajan:
> The flip in the y axis can be done with the above code and is done in the example.
Yup exactly,

I am talking about this: 
```julia
#Convert to Points
planet1_point =
    Point.((100 .* getindex.(planet1_pos, 1)), ((100 .* getindex.(planet1_pos, 2) .- 50)))
planet2_point =
    Point.((100 .* getindex.(planet2_pos, 1)), ((100 .* getindex.(planet2_pos, 2) .+ 50)))
```
Message link: https://julialang.zulipchat.com/#narrow/stream/253420-javis.2Ejl/topic/Using.20Javis.20to.20animate.20solution.20from.20DifferentialEquations/near/244088331

That `50` was half of initial `Point` .

TheCedarPrince:
So, I think what you are referring to @Ved-Mahajan is actually a different issue from what @Wikunia has in mind.

Based on my conversation with Ole on this functionality, what it assists with is taking a region of a canvas (could be the entire canvas) and maps given Cartesian points to appropriate locations on the canvas in Luxor coordinates. The user can specify where points should go in the Cartesian sense and Javis will place them where the user expects them to be by conducting an appropriate scaling. Using this scaling methodology, other scaling patterns could be applied.

What was strange about the problem you ran into, Ved, is that for some reason, the locations of the points were somehow shifted. However, I believe based on Ole's implementation here, scaling should automatically take care of this shift.

Also, @Wikunia, I just checked out and tested this branch. Shouldn't the final point here scale to `Point(-50, 100)`? I might be getting Luxor mapping flipped in my head - feature testing at 2am does not help either! :joy: 

```julia
julia> scale_point = scale_linear(Point(-10, -10), Point(10, 10), Point(-100, -100), Point(100, 100))
Javis.LinearScale{Point}(Point(-10.0, -10.0), Point(10.0, 10.0), Point(-100.0, -100.0), Point(100.0, 100.0), true)

julia> scale_point(Point(5, 5))
Point(50.0, 50.0)

julia> scale_point(Point(-5, 5))
Point(-50.0, 50.0)

julia> scale_point(Point(-5, -5))
Point(-50.0, -50.0) # Is this wrong? 
```

Wikunia:
Thanks for the the review. In your case the signs on the from rectangle (the first two points in the `scale_linear` function) are the same as those of the to rectangle (the last two points in the `scale_linear` function) that means applying it to one point will never flip the sign which it doesn't. If you want to flip the y axis sign as Luxor does you have to define the "to rectangle" in such a way that it is flipped there.

Ved-Mahajan:
> What was strange about the problem you ran into, Ved, is that for some reason, the locations of the points were somehow shifted. However, I believe based on Ole's implementation here, scaling should automatically take care of this shift.

Yeah exactly, the reason for that is unknown to me too. Basically what you suggested as a solution is kind of the same mapping as what Ole is doing `x -> scaling_factor*x + value_to_shift_by`. In that case, the `scaling_factor` was `100` and `value_to_shift_by` was `50` which is exacty **half of the scaling factor**. 
There was one more question to this, the initial point was `(1,0)` so with the scaling factor without shift it would be `Point(100,0)` and you add `Point(50,0)` to it. DO you notice that shifting factor is, coincidentally, **half of initial Point after scalinng up!** 
I am not very sure about why this happened lol.

But yeah ,all in all, this linear maping seems really exciting as it is going to save trouble of doing all the conversion manually!

PR #413, opened 2021-09-11
Ignore the RFrames for the first Action

### LINKED ISSUE #335: [BUG] Frame range for first action out of bounds when using RFrames
**Describe the bug**
When using `RFrames` on the first action of the object the frame range to it gets assigned beyond the object's own global frame range. This happens due to the value of `last_frames` being equal to the parent objects global frame range at the time of calling [get_frames](https://github.com/Wikunia/Javis.jl/blob/4241ea11c78e1e0ce6a96d3b6eae3520d36369d8/src/structs/Frames.jl#L72). Not sure if this is a bug or something that is not the correct workflow.

**To Reproduce**

1. Julia Version (i.e. output of `julia -v`): 1.5.3

2. Operating system (Mac, Linux, Windows): Mac

3. Javis version (i.e output of `] status Javis` in the REPL): Javis v0.5.0

```julia
function dummy(args...) end
video = Video(10, 10)
Background(1:20, dummy)
a = Object(3:15, (args...) -> O) # 3:15
act!(a, Action(RFrames(1:3), appear(:fade))) # 16:19
render(video; tempdirectory = "images", pathname = "")
a.actions[1].frames.frames
```

I think it is unnecessary to ruin the rest of the action frame ranges due to the first action. It would be better to treat `RFrames(1:3)` just as `1:3` for the first action using the `is_first` flag.

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #335 

When the first Action of an Object uses `RFrames` it will be ignored as it is unexpected to error of being outside the frame range.

### DIFF
--- a/src/structs/Frames.jl
+++ b/src/structs/Frames.jl
@@ -66,7 +66,7 @@ function get_frames(parent, elem, frames::Symbol, last_frames::UnitRange; is_fir
 end
 
 """
-    get_frames(parent, elem, relative::RFrames, last_frames::UnitRange; is_first=falsee)
+    get_frames(parent, elem, relative::RFrames, last_frames::UnitRange; is_first=false)
 
 Return the frames based on a relative frames [`RFrames`](@ref) object and the `last_frames`.
 """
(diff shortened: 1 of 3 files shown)


TheCedarPrince (approved):
LGTM - feel free to merge @Wikunia

PR #418, opened 2021-09-14
Fix actions on layers

### LINKED ISSUE #417: [BUG] @JLayer changes Action behaviour.
**Describe the bug**

@JLayers changes an Action behaviour it shouldn't.

**To Reproduce**

1. Julia Version (i.e. output of `julia -v`): 1.6.2

2. Operating system (Mac, Linux, Windows): Windows 10

3. Javis version (i.e output of `] status Javis` in the REPL) 0.6.2

```julia
using Javis

nolayer_video = Video(500, 500)

function ground(args...)
    background("black") # canvas background
    sethue("white") # pen color
end

Background(1:140, ground)

circ = Object(JCircle(O, 20, action=:fill, color="white"))

act!(circ, Action(1:70, anim_translate(Point(150, 150))))
act!(circ, Action(70:140, anim_translate(Point(-150, -150))))

render(nolayer_video, pathname="mwe1.gif")


layer_video = Video(500, 500)

Background(1:140, ground)

l1 = @JLayer 1:140 begin
    Object(JCircle(O, 20, action=:fill, color="white"))
end

act!(l1, Action(1:70, anim_translate(Point(150, 150))))
act!(l1, Action(70:140, anim_translate(Point(-150, -150))))

render(layer_video, pathname="mwe2.gif")
```

mwe1.gif:
mwe1

mwe2.gif:
mwe2

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #417 

Now the action.func on layers doesn't set the position but adds to it and at the end of `get_javis_frame` I tried to reset the frames positions so that the action at the next frame add to the original position rather than the modified one. Added test derived from #417 where I check that the gif's frames with and without layer are the same.


src/Javis.jl
```diff
@@ -526,9 +526,10 @@ function get_javis_frame(video, objects, frame; layers = Layer[])
 
     # check if any layers have been defined
     if !isempty(layers)
-
+        starting_positions = []
```
Wikunia, line 529: Probably `Point[]` would be better here to have the type.

test/animations.jl
```diff
@@ -715,3 +715,28 @@ end
...
+    
+    layer_video = Video(500, 500)
+    Background(1:n_frames, ground)
+    l1 = @JLayer 1:n_frames begin
+        Object(JCircle(O, 20, action=:fill, color="white"))
+    end
+    act_all!(l1)
+    render(layer_video, pathname="mwe2.gif", liveview=true)
+    
+    @test all(layer_video.layers[1].frames.frames .== nolayer_video.background_frames)
```
Wikunia, line 741: Here we would need test to compare that the gifs are actually the same.

test/animations.jl
```diff
@@ -715,3 +715,52 @@ end
...
+
+    layer_video = Video(500, 500)
+    Background(1:n_frames, ground)
+    l1 = @JLayer 1:n_frames begin
+        Object(JCircle(O, 20, action = :fill, color = "white"))
+    end
+    act_all!(l1)
+    render(layer_video, tempdirectory = "images/with_layer", pathname = "")
+
+    @test_reference "images/with_layer/0000000008.png" load(
```
Wikunia, line 740: do you mind looping over them instead? :wink:

src/Javis.jl
```diff
@@ -550,6 +551,9 @@ function get_javis_frame(video, objects, frame; layers = Layer[])
         end
 
         img_layers = place_layers(video, layers, frame)
+        for (position, layer) in zip(starting_positions, layers)
```
Wikunia, line 554: Maybe just a comment 
`# resetting layer positions to originals` or something like that

Wikunia (commented):
LGTM. I think a test case in which two actions are applied at the same time would be good too

test/animations.jl
gpucce: sorry I don't understand what I should loop the tests? I add now the actions at the same time though it look like if I try that with anim_rotate_around it doesn't work, with anim_rotate it does

Wikunia: Just 
```julia
for frame in [8, 16, 32, 64, 128, 256]
    png_name = lpad(string(frame), 10, "0")
    @test_reference "images/with_layer/$png_name.png" load(
        "images/without_layer/$png_name.png",
    )
end
```
That should work I hope

gpucce: sure will do didn't understand

Wikunia:
Great I think now it only needs a test case where `anim_translate` is applied in the same time frames like moving diagonally by moving it upward in one action and right in another action

gpucce:
> Great I think now it only needs a test case where `anim_translate` is applied in the same time frames like moving diagonally by moving it upward in one action and right in another action

aaaah I added move and rotate do I add also the double translate?

gpucce:
> > Great I think now it only needs a test case where `anim_translate` is applied in the same time frames like moving diagonally by moving it upward in one action and right in another action
> 
> aaaah I added move and rotate do I add also the double translate?

doing it now should be fast

Wikunia:
Yeah we want to check that the adding works with the position and the rotate doesn't change that or does it? :thinking:

gpucce:
> Yeah we want to check that the adding works with the position and the rotate doesn't change that or does it? 🤔

it makes perfect sense didn't think about it at first

Wikunia:
Awesome one very last thing: One bugfix line in the changelog would be good

Wikunia:
Thanks and sorry I have no idea what I did with git and the master branch... Somehow your tutorial is now in this branch which is really weird.

gpucce:
I think I added one useless commit which I then tried to revert might be I had carried more stuff but didn't realize

Wikunia:
I'm pretty sure it's my weird force push thing. I changed master back now but somehow it is still in this PR. However merging will probably fix this. I'll just let the test cases run again and merge.

gpucce:
Great! Thanks for the step by step supervision!!

Wikunia:
You're very welcome. Thanks for the fast fix :)

Issue #302, opened 2020-12-14
[BUG] Fixing the grid function


Sov-trotter:
**Before Filing a Report**

- Did I read the Documentation and/or docstrings?
- Did I search the Javis GitHub to see if this bug has already been reported?
- Am I familiar with the Javis philosophy?
- Do I have the latest version of Javis installed? 

**Describe the bug**
Cross posting here from the conversation on zulip https://julialang.zulipchat.com/#narrow/stream/253420-javis/topic/Keep.20the.20last.20frame.20of.20a.20Grid 

**To Reproduce**

1. Julia Version (i.e. output of `julia -v`):

2. Operating system (Mac, Linux, Windows):

3. Javis version (i.e output of `] status Javis` in the REPL)

4. Minimum working code example that led to bug:

**Expected Behavior and Actual Behavior**

A clear and concise description of what you expected to happen followed up with an explanation of what actually happened.

**Stacktrace (If Applicable)**
If the stacktrace includes some ffmpeg error please set the kwarg `ffmpeg_loglevel` to `"info"` i.e `render(your_video, "your_animation.gif", ffmpeg_loglevel = "info")`

**Screenshots**
If applicable, add your gif or drawing to help explain your problem.

**Additional context**

Wikunia:
This is more or less described in #39

PR #422, opened 2021-09-18
Feature grid lines aka cartesian coordinate system

### LINKED ISSUE #38: Improving Grid Line Animations

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #38 #302

One can now draw a grid given the points for the x and y axis. It's also possible to animate the creation of the grid but it's more an object that is just there now instead of everything build for animation like the current `draw_grid` function. I also removed the `draw_grid` function as I don't think it got used as it wasn't possible to have it in an non animated fashion. 

An example of using the new `coordinate_system` function:

```julia
cs = coordinate_system(Point(-100, 0), Point(200, 0), Point(0, 100), Point(0, -200); 
                    fct=arrow)
cs_obj = Object(1:nframes, cs())

act!(cs_obj, Action(1:30, appear(cs, :top_left)))
```

One thing that is a bit not so user friendly is that one needs to both have a variable for the coordinate system `cs` as well as for the object itself `cs_obj` to be able to use it in `appear`. 

The PR includes a new folder and file `objects/CoordinateSystem.jl` and it can be template for similar objects which are of use for several people. Maybe it's something @gpucce is interested in as well for a JUtitls kind of thing. 
It uses a callable struct which calls `@JShape` but saves all the information in a struct which can be then used for animating it. For this you can have a look at the new `appear` function which is also in the `objects/CoordinateSystem.jl` file.

@TheCedarPrince let me know what you think about the general idea and I can add some more features like `disappear` at least but maybe also drawing ticks on the grid. I think a combination together with #412 would be helpful which I can add when one of them gets merged. 

As an example for that: Maybe define the coordinate system and pass it in as the 3rd argument of `scale_linear` which can replace the current 3rd and fourth argument. The other way around might also make sense as then the step size for the grid could be calculated such that there is a line for each increase by 1 in the coordinate system one wants to work with using `scale_linear`.


Wikunia:
For anyone subscribed: I've added an example as the main comment. I think it makes sense as a next step to use the `scale_linear` mapping as an input to the `coordinate_system` as this could make grid lines easier to map. 
Maybe it would be nice to define the scale and use that already as the attributes for the coordinate system such that the definition of the coordinate system is done with input coords and not with canvas coords.

PR #424, opened 2021-09-19
Add postprocessing

### LINKED ISSUE #419: Add option to reverse animation
I don't know if I can't find it, but it looks like there is no easy way to reverse an animation

Maybe one could add an argument to the render function so it is changes from 
```
function render(
    video::Video;
    framerate = 30,
    liveview = false,
    streamconfig::Union{StreamConfig,Nothing} = nothing,
    tempdirectory = "",
    ffmpeg_loglevel = "panic",
    rescale_factor = 1.0,
)
    layers = video.layers
    objects = video.objects
    frames = preprocess_frames!(video)
```
to
```julia
function render(
    video::Video;
    framerate = 30,
    liveview = false,
    streamconfig::Union{StreamConfig,Nothing} = nothing,
    tempdirectory = "",
    ffmpeg_loglevel = "panic",
    rescale_factor = 1.0,
    reverse_animation = false
)
    layers = video.layers
    objects = video.objects
    frames = preprocess_frames!(video)
    if reverse_animation 
        frames = reverse(frames)
    end
```

One could also create a `reverse_render` function, I have no clue which is better.

I can make the PR if it makes sense.

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #419 

I add a postprocess argument to render which is supposed to be a function that gets called on a vector of frames as returned by `get_javis_frame`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Javis.jl - Changelog
-- Added fix to `latex` function to make it work better on Windows.
+
+## v0.6.4 (19th of September 2021)
+- Added fix to `latex` function to make it work on Windows
 
 ## v0.6.3 (17th of September 2021)
 - `RFrames` is ignored when used in the first `Action` of an `Object`
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Javis"
 uuid = "78b212ba-a7f9-42d4-b726-60726080707e"
 authors = ["Ole Kröger <o.kroeger@opensourc.es>", "Jacob Zelko <jacobszelko@gmail.com>"]
-version = "0.6.3"
+version = "0.6.4"
 
 [deps]
 Animations = "27a7e980-b3e6-11e9-2bcd-0b925532e340"


gpucce:
@Wikunia I will move on to add tests. While I do that can you check if it makes sense the way I made it? Thanks!

src/Javis.jl
```diff
@@ -275,15 +276,35 @@ function render(
         tempdirectory = mktempdir()
     end
 
+    unprocessed_frames = Vector{Matrix{RGB{Normed{UInt8,8}}}}()
+
     filecounter = 1
-    @showprogress 1 "Rendering frames..." for frame in frames
+    @showprogress 1 "Drawing frames..." for frame in frames
         frame_image = convert.(RGB, get_javis_frame(video, objects, frame; layers = layers))
+        push!(unprocessed_frames, frame_image)
```
Wikunia, line 284: I think saving them all into a vector might slow down the render function dramatically.
I'm wondering whether it makes more sense to call `postprocess` with the name of the tempdirectory folder instead. 
That would only work for `gif` though. Another option would maybe be to check if the `postprocess` function is the identity function. If that is the case then do nothing otherwise use the workflow you provided. However for simple postprocessing where one does a change for each frame and don't need all of them this current way would still potentially slow it down/fill up the memory.

Wikunia:
Also make sure to merge the latest master into this as there were some changes to the `render` function.

src/Javis.jl
gpucce: Ok I thought the vector would be a catch but don't really know how to work around that. Let me try a bit more if I don't come up with anything I add it only as a frame by frame function and then for the reversing I add it as a special case with its own keyword argument maybe?

Wikunia: Yeah the problem is that all the images need to be in memory for this. Which is probably fine for small short animations but for people who want to create YT videos with it it's probably not going to end up well :smile: 
Special case for reverse seems reasonable to me.

PR #427, opened 2021-09-21
Add postprocessing take 3

### LINKED ISSUE #419: Add option to reverse animation

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #419 

I add 2 keyword arguments to the `render` function, `postprocess_frames_flow` and `postprocess_frame`. 

The first is called on the `frames` vector and can be used to extend it and manipulate it however one wants. 

The second is called after each `get_javis_frame` call and adds a last postprocessing step. In order to not repeat this call in those cases when the same frame appears more than once in the frames vector I try with this: add a `frames_memory` Dict to act as a memoization for the frames but at each iteration check if the current frame will appear in the remainder of the frames to process and in case it does not remove it from the memory.


gpucce:
Hi @Wikunia , I tried to do what we said on zulip, that is:
 - move postprocessing in a function
 - move rescaling within the postprocessing function
 - add docstrings to all functions and comments
 - add tests
Let me know what is missing and I will try to add it!

src/Javis.jl
```diff
@@ -233,6 +233,12 @@ Streaming to Twitch or other platforms are not yet supported.
     - Can be used if there are errors with ffmpeg. Defaults to panic:
     All other options are described here: https://ffmpeg.org/ffmpeg.html
 - `rescale_factor::Float64` factor to which the frames should be rescaled for faster rendering
+- `postprocess_frame::Function` function that is applied to the imagematrix of each frame after they have been computed 
+takes as argument postprocess_frame(frame_image, frame, frames), useful to apply a postprocessing e.g. blur to all or some
```
Wikunia, line 237: I would write `frame_image, frame, frames` so without the function name and in the code ticks

src/util.jl
```diff
@@ -260,3 +260,136 @@ Currently only implemented is to change from `<:Integer` to `float`
...
+        else
+            frame_image = _convert_and_rescale_if_needed(
+                get_javis_frame(video, objects, frame; layers = layers),
+                rescale_factor,
+            )
+        end
+    end
+
+    # Memoization container updating with cleaning
+    if !(frame in frames[(filecounter + 1):end]) & haskey(frames_memory, frame)
```
Wikunia, line 382: We normally use the short circuit operators `&&` and `||`

src/util.jl
```diff
@@ -260,3 +260,136 @@ Currently only implemented is to change from `<:Integer` to `float`
...
+        (sizemargin ÷ 2 + 2):(sizefrom - sizemargin ÷ 2)
+    end
+end
+
+"""
+    crop(im, heightto, widthto)
+
+Crops an imagematrix to size `(heightto, widthto)` around the center of the iamge.
+"""
+function crop(im, heightto, widthto)
```
Wikunia, line 287: Cropping doesn't seem to be tested.

src/util.jl
```diff
@@ -260,3 +260,136 @@ Currently only implemented is to change from `<:Integer` to `float`
...
+        (sizemargin ÷ 2 + 1):(sizefrom - sizemargin ÷ 2)
+    else
+        (sizemargin ÷ 2 + 2):(sizefrom - sizemargin ÷ 2)
+    end
+end
+
+"""
+    crop(im, heightto, widthto)
+
+Crops an imagematrix to size `(heightto, widthto)` around the center of the iamge.
```
Wikunia, line 285: *image

src/util.jl
```diff
@@ -260,3 +260,136 @@ Currently only implemented is to change from `<:Integer` to `float`
 """
 interpolateable(x::AbstractVector) = x
 interpolateable(x::AbstractVector{<:Integer}) = float.(x)
+
+
+
+"""
+    _get_range(sizemargin, sizefrom)
+
+For even numbers returns a range with values from `sizemargin ÷ 2 + 1` to `sizefrom - sizemargin ÷ 2`.
```
Wikunia, line 269: For an even `sizemargin` that is, right?

src/Javis.jl
```diff
@@ -272,20 +280,48 @@ function render(
...
+            frames = frames,
+            rescale_factor = rescale_factor,
+        )
 
         if !isempty(tempdirectory)
             Images.save("$(tempdirectory)/$(lpad(filecounter, 10, "0")).png", frame_image)
         end
         if render_mp4
-            if frame == first(frames)
+            if frame == first(frames) & !started_encoding_video
```
Wikunia, line 323: here also the short circuit `&&`

src/Javis.jl
```diff
@@ -272,20 +280,48 @@ function render(
...
+    # Create a template useful in case of postprocessing
+    frame_template = _convert_and_rescale_if_needed(
+        get_javis_frame(video, objects, first(frames); layers = layers),
+        rescale_factor,
+    )
+
+    # Check if postprocess_frames_flow is ill defined
+    original_n_frames = length(frames)
+    frames = postprocess_frames_flow(frames)
+    if !issubset(unique(frames), 1:original_n_frames)
```
Wikunia, line 292: the `unique` isn't necessary

src/Javis.jl
Wikunia: I'm unsure why we check for both here actually. Can't we just check it with `if !started_encoding_video` ?

src/Javis.jl
```diff
@@ -272,20 +280,48 @@ function render(
...
+        error(
```
Wikunia, line 293: Should be also tested with `@test_throws`

src/Javis.jl
Wikunia: Also just for good measure maybe add a test case where the first frame is not `1` which should make a difference.

Wikunia (changes requested):
Thanks again for the PR @gpucce !
I had some time for a more in depth review. Just small things and some test cases are missing, mainly for cropping.

src/Javis.jl
gpucce: I will add the test! The reason for the check is that if `postprocess_frames_flow` repeats the first frame then the encoding starts twice which I though would be bad, though I don't know VideoIO very much (at all to be fair xD)

gpucce:
I should have done it all. I could only test the returned size for the `crop` function because if I try to make two identical imagematrices one making a small imagematrix and one cropping a large imagematrix they don't match exactly. Though they do look the way they should so maybe this is fine?

src/Javis.jl
Wikunia: Yeah only checking for the first frame will fail bit only checking whether we initialized rendering should be enough, or not?

gpucce: For sure I am not understanding smth, but if after `postprocessing_frames_flow` the vector `frames` is [1,2,1] for example would't it start rendering a second time at the third step without the second check?

Wikunia: I would only use the second check 🤣

gpucce: ooooooh got I make it now xD

gpucce:
@Wikunia I did the @btime test and looks like nothing changes!!

Wikunia (approved):
LGTM. Thank you!

Sov-trotter:
This looks cool. 
I didn't take a detailed look at the code but seems like the tests compares the rendered images with itself. There is no reference image. https://github.com/Wikunia/Javis.jl/pull/427/files#diff-9e0faea70cdac333b698a72a30357133df0d1b0bbfcd8251172efecd452a47b2R42-R44

Wikunia:
It tests it with post processing and without postprocessing. So it kind of generates its own reference images all the time. Which is quite reasonable I think for seeing when things should behave exactly like a different version of it.

Sov-trotter:
Ah. I saw that now. Sorry

PR #431, opened 2021-10-02
Postprocessing frame check bug fix 

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #

The check for frames returned by `postprocess_frames_flow` was wrongly written, this should fix it.


Wikunia:
I don't know how the previous code could result in a failure. Probably missing something here. Aren't the frames always from 1-... ? Isn't gif rendering would fail otherwise? Maybe add a test which would fail in the current version and is fixed by this.

Btw make sure to update your personal fork before creating a PR. (For the next one don't worry about this one) as currently this PR has >30 commits but only the last is the one which isn't merged already

gpucce:
Sorry for merging I though I did. 

The error was this, I think for javis is not an issue having for example frame 0 defined but with the old solution this would throw this error instead of using frame 0 normally, now it shouldn't anymore.

gpucce:
So I adapted a test in postprocessing to account for the negative frames and it does work with this PR while it does not without it.

Wikunia:
Great thanks

Wikunia (commented):
Thanks LGTM. Please add a line to the changelog :) 
Then I'll merge and tag a new release

gpucce:
That should be it!

PR #434, opened 2021-10-20
first attempt at DelayedPosition

### LINKED ISSUE #433: add startpos method
It is generally hard to make a translation from an user-unknow (but javis-known) current object position to a specific point.

Talking with @Wikunia he came up with a possible interface `startpos` that should be a function that when called at non-render time returns `nothing`. The first time it is called at render time returns an object current position, whereas when called again at render time it keeps returning the position of the object at the time it was originally called.

The only alternative I can come up with is a custom `action.metafunc` that is called before `action.func` and allows for user defined solutions, though this is really not user-friendly.

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #433 

I added a new struct called `DelayedPosition` a new constant `const STARTED_RENDERING = [false]` and function `get_delayed_position` which is almost a `DelayedPosition` constructor.

Then I added a method to `get_position(dp::DelayedPosition)` that checks if the rendering has started, using the `STARTED_RENDERING` constant and then if it is called for the first time sets `dp.position = get_position(dp.obj)` and returns `dp.position`, otherwise just returns `dp.position`.

all together this should allow smth like this 

```
testvideo = Video(200, 200)
Background(1:30, (args...) -> begin; background("white"); sethue("black") end)
ball = Object(JCircle(Point(-50, 0), 10, color="black", action=:fill))
act!(ball, Action(anim_translate(Point(-50, 0), Point(50, 0))))
act!(ball, Action(15:30, anim_translate(Javis.delayed_pos(ball), Point(0, -50))))
render(testvideo)
```

to return smth like this where the second upward shift starts from the circle position at the frame when the action with `get_delayed_position` starts:

test

**Problems**
I had to add DelayedPosition amone the field types of `Translation` and among the argument types of `anim_translate` and this might be needed else where I still don't know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ##for PR
 - change font scaling in svg2luxor from 1/2 to 425/1000
 - change default line width to 2.0 like luxor 
+- add `get_delayed_position` and `delayed_pos`
 
 ## v0.8.0 (1st of February 2022)
 - Allow Luxor v3.0


src/Javis.jl
```diff
@@ -253,6 +259,8 @@ function render(
     postprocess_frames_flow = identity,
     postprocess_frame = default_postprocess,
 )
+
+    STARTED_RENDERING[1] = true
```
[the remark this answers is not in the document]
gpucce: - changed STARTED_RENDERING to CURRENTLY_RENDERING

- didn't move `CURRENTLY_RENDERING[1] = false` inside `empty_CURRENT_variables` because the latter is not called if `pathname` is empty, I don't know if there is a reason for that. if not I can move it inside it and make it a bit cleaner.

src/structs/Transitions.jl
```diff
@@ -5,8 +5,8 @@
 =#
 
 struct Translation <: AbstractTransition
-    from::Union{Object,Point}
-    to::Union{Object,Point}
+    from::Union{Object,Point,DelayedPosition}
```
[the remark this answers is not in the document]
gpucce: This one I don't fully understand

gpucce:
@Wikunia I don't know what happened with codecov but I think it is disallowed on juliaanimators or smth like that. I think I sent you and @TheCedarPrince a the authorization to enable it.

gpucce:
Thanks!

Wikunia:
Codecov should be activated now. Feel free to go back to draft mode (I clicked the wrong button but unsure how to go back :smile: )

Wikunia:
Hope there will be the codecov checks when you push something here again. If not ping me again and I'll check whether there are some other tokens I need to add or something like that.

gpucce:
There is one issue to look into, @Wikunia I'd appreciate some comments here because it is a matter of how one likes I think. Basically, since in `_anim_...` functions there is a line similar to `p = p - obj.start_pos` as it is, `delayed_pos` behaves "differently" from a user perspective, whether one:
- acts on `obj1` using `delayed_pos(obj2)`
- acts on `obj1` using `delayed_pos(obj1)`

Now, one can make it work as one expects in only one of the two scenarios, unless maybe one adds a sort of `GAction` that is an `Action` that does not account for relative positions. Anyway, besides random solutions, some idea would be helpful because other than this I think by now the delayed idea itself kind of works.

gpucce:
@Wikunia the codecov still seems to not show up, anyway now I am  stopping so no rush in checking it, thanks!

Wikunia:
Okay i'll get back to you. In general based on our discussion before, I would assume that only one of the cases which you currently have in your example to work. If it's computes more or less random results for the other case, I assume it's fine if there's no other choice that is more reasonable.

Wikunia:
codecov should work in general though based on #441

gpucce:
> codecov should work in general though based on #441

True, do you have any idea why it does not show among the checks here? It might be the way it should be and I just don’t get why, I seem to remember that it would run all the times normally

Wikunia:
I don't know this one was transferred from the old repo to the new whereas #441 was created here but I don't know why this should make a difference :smile: 
You can have a look here to check your codecov though: https://app.codecov.io/gh/JuliaAnimators/Javis.jl/branch/add_delayedposition

gpucce:
> I don't know this one was transferred from the old repo to the new whereas #441 was created here but I don't know why this should make a difference 😄
> You can have a look here to check your codecov though: https://app.codecov.io/gh/JuliaAnimators/Javis.jl/branch/add_delayedposition

Great could find it, looks like coverage is doing almost perfectly!! I will add a few more then make it ready for review!

Wikunia:
Regarding your question from yesterday, I would say it only makes sense when in `anim_translate` the starting position has to be the position of the actual object at that moment.
Because otherwise it doesn't also make sense without delayed position. At the moment, it is the case that the starting position always has to be the position that the object is at that point. For example, in your example from above, it's the case that you specify the starting position of the ball once and then the starting position of the ball and at that time in `anim_translate`. If you change the y position of the start in the translate function without the delayed position, it also is messed up which I think is reasonable because that's why we introduce the starting position of that function in the first place.

Wikunia:
Also maybe you can add a more reasonable real world example than the one at the beginning of the PR to make clear why this is needed.
The current one can simply be written as:
```
testvideo = Video(200, 200)
Background(1:50, (args...) -> begin; background("white"); sethue("black") end)
ball = Object(JCircle(Point(-50, 0), 10, color="red", action=:fill))
act!(ball, Action(anim_translate(Point(-50, 0), Point(50, 0))))
act!(ball, Action(25:50, anim_translate(Point(0, -50))))
render(testvideo, pathname="test.gif")
```
(I change the frame length but that doesn't matter :smile: )

Wikunia:
I mean it's a good use case just the position is chosen in such a way that it doesn't show the power :smile: 
Because at that point the ball is at the origin. We might even want to have `anim_translate_to(P)` which calls `anim_translate(delated_pos(obj), P)` to simplify this.

gpucce:
Sure I Will add the nicer use case also in the how to section of documentation, also I can the anim_translate_to in this same pr

Wikunia:
Maybe check if it's really as simple because `anim_translate_to` doesn't have the information of the object it is getting applied to at that stage. We would have to think about a way to apply the `delayed_pos` to the right object.
If it's not really simple I would prefer to make another PR to avoid making this one longer than necessary.

gpucce:
> Maybe check if it's really as simple because `anim_translate_to` doesn't have the information of the object it is getting applied to at that stage. We would have to think about a way to apply the `delayed_pos` to the right object. If it's not really simple I would prefer to make another PR to avoid making this one longer than necessary.

You are right I will add it later, I will add the more meaningful use case as soon as I can!

Wikunia:
@gpucce would you mind to remind me what needs to be done for this? More test cases or also some functionality/bugfixes?

gpucce:
> @gpucce would you mind to remind me what needs to be done for this? More test cases or also some functionality/bugfixes?

Hi @Wikunia I'm really sorry I have not finished it yet but I am working more than I actually can. 
It should mostly be a couple test cases and docs, plus the example in the how-to section. 

I will try to work on it tomorrow morning

Wikunia:
Hi @gpucce no worries as you can see I also didn't finish up what I was working on due to my payed work 😅 Just wanted to catch up. So no hurry and let me know if I can help you with anything

src/structs/Transitions.jl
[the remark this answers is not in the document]
Wikunia: We could have `Translation{F,T}` and then 
```
from::F
to::T
```
don't know whether that is cleaner though. Let's leave it like that for now.

src/Javis.jl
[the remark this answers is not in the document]
Wikunia: I think we should call it in all cases.

gpucce:
@Wikunia as I told you on zulip I went on and added support for all shorthands except for @JShape but there is a way to do it i will add it some where.

However now we can do this 

```
vid = Video(500, 500)
Background(1:200, (args...)->background("black"))
o1 = Object(JCircle(O, 30, color="red", action=:fill))
act!(o1, Action(anim_translate(Point(200, 0))))
o2 = Object(50:200, JStar(Javis.delayed_pos(o1), 20, color="blue", action=:fill))
act!(o2, Action(anim_translate(Point(0, 100))))
o3 = Object(100:200, JBox(Javis.delayed_pos(o2), 10, 10, color="orange", action=:fill))
	
Object(100:150, JLine(Javis.delayed_pos(o2), Javis.delayed_pos(o1), 
	color="white"))
act!(o3, Action(1:50, anim_translate(Javis.delayed_pos(o2), Javis.delayed_pos(o1) + Point(10, 10)))) 
# Nore that you can add point to delayed_pos :) 
render(vid, pathname="test_vid.gif")
```

And the result is this 

test_vid

Wikunia:
Awesome! Thanks a lot. One small question: what exactly does the last action mean? It's always hard to reason about an anim translate when the starting position is not the position of the object itself, right?

gpucce:
In this case it is the same position maybe? 

In general it will work as if you had replaced the two points with the points where the other objects are at the time when the Animation starts, hopefully

Wikunia:
I was wondering as the end position should be at the right of the circle center. It just takes the difference of end and start point of the two given which makes it hard to reason about but I think we don't need to worry about that here. Would just suggest that you use something else for test cases and documentation where it is clearer what should happen 😊

gpucce:
Sure I will, though also in this case looks like it behaves the way it should, if you wonder why it does not reach the end of the white line exactly is because I added `+ Point(10, 10)` in the action just yo show that it is possible!!

Wikunia:
But o1 is the circle isn't it ? 🤣

gpucce:
Yeah! But it will aim for the position of the circle at moment the yellow brick starts moving, this si the behaviour in general

gpucce:
Hi @Wikunia it would be nice to have your review!! 

I'll add the test to raise the codecov but there should be nothing really important missing there for you to review the PR!

Wikunia:
Currently on the move. Will check later today 🙂

docs/src/howto.md
```diff
@@ -212,9 +220,50 @@ The live viewer can be called with adding `; liveview=true` to the [`render`](@r
...
+
+## Why does `pos` always throws error?
+
+Both `get_position` and `pos` are a bit unintuitive in the sense that one may think that this should work:
+
+```julia
+vid = Video(500, 500)
+Background(1:50, (args...)->background("black"))
+o1 = Object(JCircle(O, 10, action=:fill))
+render(vid)
```
Wikunia, line 238: I think you've missed a `o2` object here or something. Also maybe set the background color to white as otherwise it's a black circle on a black background

docs/src/howto.md
```diff
@@ -212,9 +220,50 @@ The live viewer can be called with adding `; liveview=true` to the [`render`](@r
...
+
+However there is an alternative to `pos` to make the first example work, `get_delayed_position` and you can use this just like above:
+
+```julia
+vid = Video(500, 500)
+Background(1:50, (args...)->background("black"))
+o1 = Object(JCircle(O, 10, action=:fill, color="red"))
+# it works with or w/o this line and the effect changes coherently!!!
+act!(o1, Action(anim_translate(Point(100, 0))))
+o2 = Object(25:50, JCircle(Javis.get_delayed_position(o1), 5, action=:fill, color="blue"))
```
Wikunia, line 267: I think we should export this amazing function :wink:

docs/src/howto.md
```diff
@@ -212,9 +220,50 @@ The live viewer can be called with adding `; liveview=true` to the [`render`](@r
...
+embed(vid)
```
Wikunia, line 268: maybe change to `render` as the other code snippets use render.

src/Javis.jl
```diff
@@ -134,6 +135,10 @@ function centered_point(pos::Point, width::Int, height::Int)
     Point(pos.x - width / 2, pos.y - height / 2)
 end
 
+function centered_point(pos::DelayedPosition, width::Int, height::Int)
```
Wikunia, line 138: A small docstring would be great. Also does width and height need to be `Int64`?

src/structs/Transitions.jl
```diff
@@ -5,8 +5,8 @@
...
+    to::Union{Object,Point,DelayedPosition}
 end
```
Wikunia, line 10: can this be `Union{Object, Luxor.AbstractPoint}` now?

test/delayed.jl
```diff
@@ -0,0 +1,204 @@
...
+    Background(1:31, (args...) -> begin
+        background("white")
+        sethue("black")
+    end)
+    ball1 = Object(1:31, JCircle(O, 10, color = "black", action = :fill), Point(-45, 0))
+    ball2 = Object(1:31, JCircle(O, 10, color = "blue", action = :fill), Point(0, 0))
+    act!(ball1, Action(1:31, anim_translate(Point(-45, 0), Point(45, 0))))
+    act!(ball2, Action(16:31, anim_translate(Point(25, 0), Point(0, -45))))
+    mkdir("images/not_delayed/")
+    render(testvideo, tempdirectory = "images/not_delayed", pathname = "")
```
Wikunia, line 14: This part doesn't seem to get tested?

Wikunia (commented):
I made some small comments. 
One question regarding the `DelayedPosition` struct.
Would it make sense to have the ability to get the x and y component with the `.x` / `.y` syntax directly? 

I think it that case we might be able to change all function arguments in Luxor to support `AbstractPoint` and we wouldn't need to call this `get_position` function for each `AbstractPoint` we send over to `Luxor`? 

I mean for now it's okay as it is and then we don't need to create another Luxor release but it might be reasonable for the future. Let me know what you think.

gpucce:
> I made some small comments. One question regarding the `DelayedPosition` struct. Would it make sense to have the ability to get the x and y component with the `.x` / `.y` syntax directly?
> 
> I think it that case we might be able to change all function arguments in Luxor to support `AbstractPoint` and we wouldn't need to call this `get_position` function for each `AbstractPoint` we send over to `Luxor`?
> 
> I mean for now it's okay as it is and then we don't need to create another Luxor release but it might be reasonable for the future. Let me know what you think.

That would be amazing

test/delayed.jl
Wikunia: Ah I see now. Makes sense.

gpucce:
@Wikunia I should have addressed all the comments!

Wikunia (approved):
LGTM you might wanna read through the documentation @TheCedarPrince 

Also make sure to mention the functionality in the changelog @gpucce.

PR #441, opened 2021-11-05
add global testset

### LINKED ISSUE #440: add global @testset
Could we add global `@testset` wrapping all `@testset` so that if some tests don't pass one gets to see how the others went anyway?

I ask because for example I work woth WSL and the liveview test can't work here so I have to add the global `@testset` locally

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #440


gpucce:
@wikunia about what you asked on #441, try to check the output of the tests to see if you like it, the fact is that if all tests go well it does not show you much, whereas if they don't then it does.

Wikunia:
I assume it shouldn't be in draft state, right? Will merge if you're done with it.

PR #442, opened 2021-11-07
return nothing if pathname is empty and empty the constants

Wikunia:
## PR Checklist

The `CURRENT` constants should always be reset even if the liveviewer is used without a pathname. 
This fixes one point mentioned by @gpucce in #434 

I return `nothing` in that case and call `empty_CURRENT_constants`


gpucce (approved):

gpucce (approved):

Wikunia:
Hmm I see. I think it's more a testing bug though but will have a look later. I think we shouldn't call `render` and assume to have access to the stuff later on. That is kind of what emptying the constants should make sure in the first place.

gpucce:
> Hmm I see. I think it's more a testing bug though but will have a look later. I think we shouldn't call `render` and assume to have access to the stuff later on. That is kind of what emptying the constants should make sure in the first place.

yeah I think I get it, it happens because in the testing there are calls to functions involving `vid` after it has already been rendered right?

Wikunia:
We should also empty those when we call the liveviewer which is done at the top of the `render` function.

gpucce (dismissed):
Did not mean to ask my own review

Wikunia:
Haha you approved it three times now and it is still buggy :smile:

gpucce:
> Haha you approved it three times now and it is still buggy 😄

I don't know how this works, I wanted to make it un-approved but apperently that is not how it works? (sorry for the mess)

Wikunia:
@TheCedarPrince I'm not super sure how the liveviewer works. Is there any possibility to clear the constants after the gtk window was closed?

gpucce (approved):

PR #443, opened 2021-11-07
make it possible to call Luxor when outside rendering

### LINKED ISSUE #438: Making Luxor work completely inside Javis
Sometimes one wants to just draw a single image inside a bigger Javis project for example inside Pluto.
Then one gets weird errors like `CURRENT_VIDEO is empty`. 

Based on my discussion with @gpucce it would be great to be able to use 
```
@png begin 

end
```
as an example outside of any animation. One probably needs to just check that one calls the Luxor functions for those that are overwritten by Javis otherwise. Which means checking if the video is empty and then calling the Luxor functions.

I'll have a look how difficult this is in the next couple of days.

Wikunia:
## PR Checklist

**Link to relevant issue(s)**
Closes #438 

I've added a `CURRENTLY_RENDERING` const same as @gpucce in #434 and check in each function in `luxor_overrides.jl` whether we are inside a rendering. If not the standard Luxor function is called. 

More checking would be good. Maybe we find some old issues where this issue was raised.


Wikunia:
Seems like there is some issue with MacOS and `text` maybe.

src/Javis.jl
```diff
@@ -1,5 +1,6 @@
 module Javis
 
+using Base: kwarg_decl
```
gpucce, line 3: Hi @Wikunia I don't know what `kwarg_decl` does but I think it is not called anywhere else so maybe it should not be there?

gpucce (approved):
Other than that small thing I think it works

src/Javis.jl
Wikunia: Oh I somehow missed your review. Yes I don't know why but when I type some weird stuff in VSCode by accident some automatism adds those Imports. Will remove it.

TheCedarPrince:
Hey @Wikunia and @gpucce !
I'd love to add my thoughts but I really am not quite sure as to what this PR is doing?
Does it just allow one to create Luxor drawings alongside Javis?
Could you maybe give me a small example here @Wikunia ?

Thanks!

~ tcp :deciduous_tree:

Wikunia:
Hey @TheCedarPrince 
currently on the move so for now only this example in our test cases: https://github.com/JuliaAnimators/Javis.jl/blob/3ff2097fb304e95838a8eccb6d921720ea4e1272/test/unit.jl#L372

The idea is to be able to play around with Luxor before going into animations in the same file. With this PR one is able to do Luxor drawings by importing Javis but then just use Luxor functions without any of Javis objects and stuff.

Hope that makes sense

Wikunia:
Would love to have this in v1.0 of Javis :wink:

src/luxor_overrides.jl
```diff
@@ -240,6 +244,14 @@ draws the text from left to right in the first 15 frames and in the last 15 fram
...
 """
 function text(str, pos = O; valign = :baseline, halign = :left, angle = 0.0)
+    call_luxor_if_outside_rendering(
+        Luxor.text,
+        str,
+        pos;
+        valign = valign,
+        halign = halign,
+        angle = angle,
+    ) && return
```
ArbitRandomUser, line 254: will this return what Luxor.text returns ? 
might not be a big issue , but if user  is relying on Luxor functions return value it would be nice to return what Luxor returns

Wikunia: No it doesn't good point will change.

Wikunia: Fixed

Sov-trotter (approved):

PR #449, opened 2021-12-25
Allow integers and irrationals in Rotation angle

leephillips:
An innocent user trying
    act!(object1, Action(anim_rotate_around(π, object2)))
will get a method error, because of the way Javis.Rotation is defined
using a parametric type. Same deal for
    anim_rotate[_around](o1, <integer>, o2)
The small change in the definition of Rotation in this commit allows
the use of Irrational and Integer angles in anim_rotate_around()
and anim_rotate().

## PR Checklist

**Link to relevant issue(s)**
Closes #

### DIFF
--- a/src/structs/Transitions.jl
+++ b/src/structs/Transitions.jl
@@ -31,9 +31,9 @@ anim_translate(x::Real, y::Real) = anim_translate(Point(x, y))
 anim_translate(tp::Point) = Translation(O, tp)
 anim_translate(fp::Union{Object,Point}, tp::Union{Object,Point}) = Translation(fp, tp)
 
-struct Rotation{T<:Real} <: AbstractTransition
-    from::T
-    to::T
+struct Rotation <: AbstractTransition
+    from::Real
+    to::Real
     center::Union{Nothing,Point,AbstractObject}
 end
 


TheCedarPrince:
Hey @leephillips, thanks for the interest in Javis!

Generally our process for adding or amending functionality to Javis is detailed in this contributing guide for Javis: https://juliaanimators.github.io/Javis.jl/stable/contributing/

Right now I do not clearly see why this is needed and also that this PR does not have any tests for this new functionality nor anything checked off from our PR checklist.
Could you open an issue and we discuss why you feel this functionality is needed there?
Otherwise we will most likely close this PR.

Thanks for the interest and we look forward to continuing the conversation!

~ tcp

leephillips:
Sorry, next time I’l open the issue first and name my branch properly. And test.

leephillips:
I tested by branch according to your contributor instructions and all tests passed:

```
Test Summary: | Pass  Total
Javis         |  532    532
     Testing Javis tests passed 
```

which was not surprising for this minor code adjustment.

Wikunia:
Thanks for opening the issue and PR. I actually encountered the error before. I think it would make Javis more user friendly if we allow integer and irrational numbers for rotations. However I don't think that this PR is the best version around it. With `Real` in the struct definition the type is not explicit anymore for future calculations. Which probably results in shower code. Though currently it probably doesn't matter. 

I think however that using `T` as before should work we just need to change some methods when constructing the `Rotatiom`. 

Regarding test sets: @TheCedarPrince meant that the new functionality isn't tested. We normally want that for new features or Bugfixes that the old code with the new test would fail bit the new voice with the new test pass. Hope that makes sense.

Thanks again for the issue and PR. If you isn't any further questions feel encouraged to ask here 😊

leephillips:
I’ll look at other ways to fix this bug. Do you mean that I should write a test for this, which should be part of the commit?

Wikunia:
Yes my idea would be to probably convert everything to the most broad type here: https://github.com/JuliaAnimators/Javis.jl/blob/86a3655fe41fba47a8eeac10572a3de917e81edf/src/structs/Transitions.jl#L58 and also in `anim_rotate_around`
Currently it says the start angle to `0.0` but should set it to `zero(T)`. Hope that makes sense. 
Otherwise I'll fix this myself tmr.

leephillips:
That does make sense. If you want to hold off I’m pretty sure I can get this completed tomorrow. I’ll close this PR now and open a new one with the fix as you suggested tomorrow (named according to your convention).

leephillips:
@Wikunia Your fix works for Integer angles but not for Irrational angles. The problem is the surprising result of zero(Irrational).

My new fix for this problem is to define

```
anim_rotate_around(ta::Real, p) = Rotation(0.0, ta, p)
anim_rotate_around(fa::Real, ta::Real, p) = Rotation(fa, ta, p)
```
in Transitions.jl, and to also define the constructor

```
Rotation(from::Real, to::Real, center) = Rotation(promote(from, to)..., center)
```

without touching the definition of the struct Rotate.

This covers all combinations of Real angles, including Integer and Irrational in particular. Tests pass.

I’ll wait for your response before writing a test and making a new PR. My commit is at https://github.com/leephillips/Javis.jl/tree/leephillips-issue-450 if you want to see it.

Thanks, 
Lee

Wikunia:
Thanks yeah that looks good to me. Thanks for testing the `zero` function. `promote` is a good workaround. 
Some test cases would be awesome.

PR #451, opened 2021-12-27
Migrate viewer functions from Javis to JavisNB

gpucce:
## PR Checklist

If you are contributing to `Javis.jl`, please make sure you are able to check off each item on this list:

- [x] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [x] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [ ] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [ ] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [ ] Did I properly add test dependencies to the `test` directory (if applicable)?
- [ ] Did I check relevant tutorials that may be affected by changes in this PR?
- [x] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**
Closes #

**How did you address these issues with this PR? What methods did you use?**
The idea is to move notebook viewers to JavisNB and change render in order to play well with the different package. 

More details in this PR https://github.com/JuliaAnimators/JavisNB/pull/6 where the parallel work on JavisNB is done.

Issue #113, opened 2020-08-21
Releases


Wikunia:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

TheCedarPrince:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

Wikunia:
@JuliaRegistrator register()

[...]

PR #458, opened 2022-01-17
Prepped for tagging a v0.8.0 release of Javis

TheCedarPrince:
## PR Checklist

- [X] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [X] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [X] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [X] Did I properly add test dependencies to the `test` directory (if applicable)?
- [X] Did I check relevant tutorials that may be affected by changes in this PR?
- [X] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**
Closes N/A

As a result of PR #451, since this is a breaking change, we should probably release a new version of Javis.
I went ahead and updated the Changelog as well as Project TOML for the release.
Any more PRs or features we want to introduce before we release a new version?
Maybe we should try to get #399 in as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Javis.jl - Changelog
 
+## v0.8.0 (17th of January 2022)
+- Migrated Javis notebook tools to JavisNB
+	- In order to use Javis within notebooks (i.e. Jupyter and Pluto notebooks), please refer to [JavisNB](https://juliaanimators.github.io/JavisNB/)
+
 ## v0.7.2 (2nd of January 2022)
 - Fix bug that would make frames <= 0 throw error
 - Allow integer and irrational angles in rotation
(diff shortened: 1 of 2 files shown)


Wikunia:
LGTM though JavisNB doesn't seemed to be tagged yet, right?
I wouldn't wait for #399 for this release

gpucce:
@Wikunia @TheCedarPrince I have never tagged a release, if we choose to go on before #399 I'd like to do it myself so I know how it works. In case it is complicated or there are things to be particularly careful about you can point me to those.

Wikunia:
This has to be installed:
https://github.com/JuliaComputing/Registrator.jl#install-registrator
I just checked and it is for all repos of this organization.

Then you need to make sure that the Project.toml is correct (has version 0.1.0, has compat bounds for every package it depends on)
Afterwards you should be able to simply open an issue called releases (actually doesn't matter how you call it)
and call `@JuliaRegistrator register()` as you can see here: #113 

Afterwards a new issue on the JuliaRegistry should be created (the bot will answer your comment in the issue where you called `register`. We need to wait for 3 days until it is registered.

Then we can continue with v0.8 here

gpucce:
Ok then tomorrow I will do it. Now it would be impossible without mistakes somewhere.

PR #461, opened 2022-01-22
move viewer to JavisViewer

### LINKED ISSUE #460: Move liveview outside Javis
Moving Liveview to a Different Package `JavisViewer`

gpucce:
## PR Checklist

**Link to relevant issue(s)**
Closes #460 

In this PR I remove all parts of `Javis` that are used for the liveview and will add them in the `JavisViewer.jl` package. 

Once I create a PR in JavisViewer I will reference this one.


gpucce:
@Wikunia @TheCedarPrince And with a new PR a new interface question. 

I am going similar to `JavisNB` with the equivalent of `embed` being `stream` the question is I am moving the `liveview=true` option into stream and entirely out of `render`, or at least that would be the plan. Is this a good idea or do you prefer the possibility to call liveview from  `render` itself? Perhaps the best is having it in both `render` and  `stream` acting in the same way?

Let me know what you think!

Wikunia:
I think I would use a `liveview` function which works similar to the `embed` function. No need for a liveviewer keyword anymore. Stream sounds more reasonable to me when we have a stream to twitch functionality or something like that.

PR #464, opened 2022-02-08
Latexbranch

ArbitRandomUser:
## PR Checklist
primarily this PR adds support for using dvisvgm which is distributed with TeXLive  to render latex . This is an alternative to the existing mathjax solution. Mathjax still exists as the default setting for now.

All tests dont pass because the images are not exact 1-1 pixel replicas .. but they are close enough 
edit1: Former are from dvisvgm and latter are from mathjax tex2svg

Linux: To get this working you will need `latex` (symlinked to pdftex binary) and `dvisvgm` in your path, pdftex, the `latex` symlink,  and dvisvgm are made available by installing TeXLive (most distributions package  TeXLive  in their repos)

To try out an example program 
https://gist.github.com/ArbitRandomUser/da8d439544945466c1df566df561d5f4

Windows & MacOS:   somebody's gotta test this for me , i dont have access to either OS's at the moment.

act_dancing_circles_16_rot_trans
ref_dancing_circles_16_rot_trans

act_ologn
ref_ologn

(by default \begin{equation} adds numbering in latex)
act_matrix_transpose1
ref_matrix_transpose1

act_latex_8
ref_latex_8

- [✔️] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [ ✔️] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [ ✔️] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [ ✔️] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?

**Link to relevant issue(s)**
Closes #


src/latex.jl
```diff
@@ -139,24 +147,89 @@ end
...
+        elseif Javis.LaTeXprog == :dvisvgm
+            println("using dvisvgm")
+            svg = tex2svg(text)
         end
         LaTeXSVG[text] = svg
     end
     return svg
 end
+
+function tex2svg(text::LaTeXString; output_dir = "./.TeXfolder")
```
Wikunia, line 181: I'm not sure this output dir works for windows systems. I think there is a Julia function to create a temporary folder.

src/latex.jl
```diff
@@ -139,24 +147,89 @@ end
...
+    """
+    generates svg from TeX;
+    default folder placed is ./.TeXfolder,
+    pdf's and aux files are deleted , tex and log files and other files remain
+    add LaTeX packages in `packages=[]`
```
Wikunia, line 186: I don't understand the comment. Is there a reason why they don't get deleted? 
Also what do you mean with `packages =[]` here?

src/latex.jl
```diff
@@ -139,24 +147,89 @@ end
...
+            end
+        elseif Javis.LaTeXprog == :dvisvgm
+            println("using dvisvgm")
+            svg = tex2svg(text)
         end
         LaTeXSVG[text] = svg
     end
     return svg
 end
+
```
Wikunia, line 180: Make sure to have doctrings here and not inside the function.

src/svg2luxor.jl
```diff
@@ -254,11 +325,15 @@ function pathsvg(svg)
     fsize = get_current_setting().fontsize
     xdoc = parse_string(svg)
     xroot = root(xdoc)
-    def_element = get_elements_by_tagname(xroot, "defs")[1]
-    # create a dict for all the definitions
     defs = Dict{String,Any}()
-    for def in collect(child_elements(def_element))
-        defs[attribute(def, "id")] = def
+    try
```
Wikunia, line 329: For each try catch I would like to know why it can fail and maybe there is an option to avoid the try catch then.

Wikunia (commented):
Thanks for opening the PR. Looks really interesting. Regarding the font sizes. Do you know why yours are bigger than before? 
I just had a look through the code so not a full review yet but maybe something of help. Will try a full review today or tomorrow. 
I think you don't have a description of how to run it with the new way yet, right? Do you mind to write down a step by step instruction in the PR for people to just try it out including how to install it. We can later put that in the docs then.

src/latex.jl
ArbitRandomUser: ah ! Pythonic habits die hard ; will fix that !

src/latex.jl
ArbitRandomUser: there's no reason to keep the pdf's around after we're done converting to svg;
i just kept it around incase people want to inspect the log and tex files if they do run into errors

ArbitRandomUser: `packages=[]` was an earlier way i was passing the packages to be added in the usepackage line of the .tex file ; ive changed it in a later commit to a module level variable `Javis.LaTeXusepackages` which is a vector holding all the package names.
will remove the `add Latex package in ...` line

src/svg2luxor.jl
ArbitRandomUser: This try statement checks if we have any `def` tags , if they exist we add the defs to  `defs` dictionary
if there are no `def` tags ( like in the case of dvisvgm which draws directly and doesnt predefine 
paths to use) `get_elements_by_tagname(xroot, "defs")` will return an empty iterable and cause a bounds error .

ArbitRandomUser: Yeah! i probably shouldn't discard all errors , should `throw(e)` if its not a `BoundsError`

ArbitRandomUser:
digging a little more deeper into the fonts it seems to me the latex-dvisvgm way is alright
mathjax imo seems to be enlarging the fonts ; maybe someone with more knowledge of fonts rendering 
can comment on this .
The earlier comment about font  exactly occupying 50 units of height when fontsize is 50 is expected behaviour
So for example the O in Olog(n) will not have to be 50 points high when the fontsize is 50.

However the O has a relative size as defined by the font, one can get this info out of a program called `fontforge`
in OTF fonts the em size or the size of the "block" is 1000 units in fontforge ;
now lets look at how many units the O glyph spans top to bottom ...
fig
725 units (shown in green ).
So after drawing all of this with an fontsize of 50 in luxor , we should expect the O's height to be 0.725*50 = 36.25 
Here i have added two dimensionings at apropriate locations from a point on y=0 to y=36.25 for one 
and y=0 to y=50 for another 

dvisvgm

We see the O is pretty close to 36.25 units , and it might be exactly 36.25 if shift the dimensioning down a little to exactly where the top of O starts , but my point is its pretty close , not by off by a visible factor like in the mathjax case.

Anyway ill try to get in the changes from the other comments done by tomorrow or day-after , maybe it can be reviewed over the weekend ? or perhaps next week if you're not free this weekend

ArbitRandomUser:
its not really mathjax's fault , it Javis scaling mathjax fonts the wrong way ,
look at the x glyph from mathjax font , its height is 425 
fig
which means the scaling in line 273 at svg2luxor.h should be 
`        scale((fsize * 425/1000) / (height / ex_height))`
instead of 
`        scale((fsize / 2) / (height / ex_height))`

Wikunia:
Thanks for checking that out ☺️

Wikunia (commented):
Let me know when you have an example for me to try it out. :wink:
I've added some small comments for things I noticed

Wikunia:
Regarding the font size I'm super happy you figured it out way more than I did :wink:

ArbitRandomUser:
@Wikunia , added an example gist in the PR :)

Wikunia:
With the gist you mentioned I'm getting the following warning:
```
┌ Warning: there maybe errors in processing latex, check /tmp/jl_EZ9e2T/jl_gTEmBu.log for details
└ @ Javis ~/Julia/Javis/src/latex.jl:209
```

with this file: https://gist.github.com/Wikunia/75b74eb6806a9fac10b52fc676c0198c not sure whether this is expected.

ArbitRandomUser:
thats a small bug ; line 208 in latex.jl should be `if !stat`;
i'm leaving `cleanup` for the temp dirs and folders as `false` in a new commit 

Also the tests fail now because ive changed scaling to 425/1000; lemme know if i
should revert it back.

PR #478, opened 2022-04-12
Fix inconsistencies in tutorial 2

mroavi:
Venus should revolve faster than the Earth around the sun.
This PR fixes this.
The frames argument was not being passed to the Venus `Object`.
An inconsistency in the text is also fixed.
The gifs have been re-generated with the fixed code.

## PR Checklist


mroavi:
I noticed that **not** passing the frames to `Object` is not an issue since it reuses the value passed to the previous `Object`. However, this is not explained in this tutorial but rather later in tutorial 3. Hence, I think it is better to pass it explicitly here. This means that the real issue of why the Earth was revolving faster than Venus was because of outdated gifs.

Ved-Mahajan:
@mroavi Hey, yeah thats correct indeed. I did not notice that while making the gif. The gif just so cool that I really did not bother to check all that!

Thanks a lot for noticing that and fixing it!!!

Wikunia:
Thanks @mroavi and sorry for the late merge.

PR #481, opened 2022-05-09
Jpoly

ArbitRandomUser:
## PR Checklist

- [  ] Did I update `CHANGELOG.md` with whatever changes/features I added with this PR?
- [  ] Did I make sure to only change the part of the file where I introduced a new change/feature?
- [  ] Did I cover all corner cases to be close to 100% test coverage (if applicable)?
- [  ] Did I properly add Javis dependencies to the `Project.toml` + set an upper bound of the dependency (if applicable)?
- [  ] Did I properly add test dependencies to the `test` directory (if applicable)?
- [  ] Did I check relevant tutorials that may be affected by changes in this PR?
- [  ] Did I clearly articulate why this PR was made the way it was and how it was made?

**Link to relevant issue(s)**

PR adds a struct JPaths .
A JPath consists of polys , fill , stroke and other meta information. 

Every Object's drawing function (i.e obj.func) is converted to an array of jpaths
at construction and stored in `obj.jpaths` . Two ways to draw the object now exist.
either call obj.func inside a Drawing() context or 
 calling drawobj_jpath(obj) , will draw the object using its jpaths instead of object.func


gpucce:
Hi @ArbitRandomUser nice having the PR, I think there is one commit, the first that is from another PR, is it intentional?

ArbitRandomUser:
yup , this builds on top of @Wikunia 's PR where you can draw with luxor without being inside Object.
this allows one to run the object.func on a drawing , which is needed to extract the path information.

PR #485, opened 2022-06-10
Render directly using FFMPEG instead of VideoIO

ArbitRandomUser:
## PR Checklist

**Link to relevant issue(s)**
Closes #

This PR gets rid of  VideoIO altogether for video rendering , some testing showed significant memory usage with using VideoIO , indicative of memory leaks .
This PR aims to fix huge RAM usage on repeatedly rendering (typically happens when using Javis in the REPL iterating changes into a video )
FFMPEG is used directly to stitch the frames together.


Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.9.0"
 Animations = "27a7e980-b3e6-11e9-2bcd-0b925532e340"
 Cairo = "159f3aea-2a34-519c-b102-8c37f9878175"
 FFMPEG = "c87230d0-a227-11e9-1b43-d7ebe4e7570a"
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
```
Wikunia, line 10: I think we can remove VideoIO here and have it as a test dependency only, right?

ArbitRandomUser: yes that should work .

src/Javis.jl
```diff
@@ -309,6 +308,10 @@ function render(
     started_encoding_video = false
 
     filecounter = 1
+    binpath = joinpath(FFMPEG.FFMPEG_jll.PATH_list[end], "ffmpeg") #is ffmpeg bin always at the end ?
```
Wikunia, line 311: can't we use `ffmpeg` like here: https://docs.juliahub.com/FFMPEG/TGvga/0.4.0/autodocs/#FFMPEG.@ffmpeg_env-Tuple%7BAny%7D

Wikunia (commented):
Besides the one comment:
I have the feeling that the memory consumption is still rising when I run something like:
```
for i in 1:100
   include("animate.jl")
end
```
for any kind of animate.jl script that renders to an mp4. In my case running `GC.gc()` reduced it back to the previous level though.

src/Javis.jl
ArbitRandomUser: yes indeed ,

ArbitRandomUser:
I did notice some memory rise on mine , but its definitely not as much as with VideoIO.jl .
it usually tends to plateau after some iterations of the loop .( and sometimes even drops down and rises again) . It might just be the GC deciding when to cleanup .

The important thing I  think the current memory is  "GC-able" . 
Unlike with  VideoIO.jl   where memory would not get cleared by the GC.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Javis.jl - Changelog
 
+# PR changes
+- changed render method for mp4 to use ffmpeg directly inplace of VideoIO
+- TODO change test to use ffprobe for  video length and we can get rid of VideoIO completely
```
Wikunia, line 5: Can you add this todo in the code and remove it from the changelog?

src/Javis.jl
```diff
@@ -752,4 +751,7 @@ export setup_stream, cancel_stream
 # scales
 export scale_linear, @scale_layer
 
+export morph
```
Wikunia, line 754: can you remove these three export lines? Seems like they are from your other branches?

src/Javis.jl
```diff
@@ -699,6 +697,7 @@ function set_object_defaults!(object)
     scaleto(desired_scale)
 end
 
+#todo add stroke/fills here
```
Wikunia, line 700: Also this doesn't seem to be related to this PR.

Wikunia (commented):
Will merge when those three things are done :wink: 
Thanks again for finding this out and coming up with this fix.

src/Javis.jl
ArbitRandomUser: remove

CHANGELOG.md
ArbitRandomUser: done

ArbitRandomUser:
Removed the one line for which we needed VideoIO as a testdep , and replaced it with a line using `ffprobe`.
So now we dont need VideoIO even as a test dependency.

ArbitRandomUser:
okay thats strange ... the tests worked on my machine , but doesnt seem to be working on github.

Wikunia:
I wouldn't worry about the test dependency.

PR #487, opened 2022-06-21
wait for ffmpeg to exit before render is complete

ArbitRandomUser:
small change , 
added `wait` for `ffmpegproc` (to exit) after we `close` it , 
Not a major issue but this ensures the `.mp4` exists after `render` is done.

A problem arises if one is going to play the video or further do some other processing on it from the script immediately after `render` .


Wikunia:
See #488

